PR #89, opened 2020-10-26
Tests for  @lowdefy/block-tools and block mocks

Gervwyk:


packages/blockTools/src/ErrorBoundry.test.js
```diff
@@ -0,0 +1,153 @@
+/*
```
SamTolmay, line 1: Typo in file name

packages/blockTools/src/ErrorBoundry.test.js
```diff
@@ -0,0 +1,153 @@
...
+  });
+  expect(comp.toJSON()).toMatchInlineSnapshot(`
+    <div>
+      one
+    </div>
+  `);
+});
+
+test('display no error message on error generated by child', () => {
+  console.error = () => undefined;
```
SamTolmay, line 45: Maybe do this only once at start, and clean up in after all?

packages/blockTools/src/ErrorBoundry.test.js
```diff
@@ -0,0 +1,153 @@
...
+  expect(comp.toJSON()).toMatchInlineSnapshot(`
+    <div>
+      Error: ErrorBoundary test error
+      <br />
+      error description
+    </div>
+  `);
+});
+
+test('display fallback component on error generated by child', () => {
```
SamTolmay, line 126: Is fallback the right name if it will override the default behavior?

yarn.lock
```diff
@@ -2837,6 +2859,15 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@lowdefy/get@npm:1.0.1":
```
SamTolmay, line 2862: Why are these here? Do we have a dep that needs to be updated?

SamTolmay (approved):

PR #116, opened 2020-11-03
blocksAntd Selector examples

ACvWyk:


packages/blocksAntd/demo/examples/Selector.yaml
```diff
@@ -1,20 +1,108 @@
...
+  properties:
+    showArrow: false
+- id: "properties.showSearch: true"
+  type: Selector
+  properties:
+    showSearch: true
+    options:
+      - Option 1
+      - Option 2
+- id: "properties.size: large"
```
Gervwyk, line 70: Let's create a test case for every value of enum?

PR #333, opened 2021-01-18
Operators

Gervwyk:


packages/operators/test/common/object.test.js
```diff
@@ -0,0 +1,237 @@
...
+    `);
+    expect(() =>
+      object({
+        params: null,
+        methodName,
+        location,
+      })
+    ).toThrowErrorMatchingInlineSnapshot(`
+      "Operator Error: _object.hasOwnProperty accepts one of the following types: array, object.
+            Received: {\\"_object.hasOwnProperty\\":null} at http://localhost/."
```
SamTolmay, line 67: Why is location = localhost here?

SamTolmay (approved):

PR #479, opened 2021-03-03
Openid Authentication

Gervwyk:
##### Todo:

- [x] _user operator.
- [x] Client access token and id token storage.
- [x] GraphQL client to send access token.
- [x] Implement setHeader on Netlify.
- [x] Read user from local store on page load.
- [x] _user on GraphQL.
- [x] Make token expiry configurable.
- [x] Unset idToken on logout.
- [x] Handle expired tokens in GQL client.
- [x] Throw when both a public and private page list is provided.
- [x] Add public or private flags to build artefacts.
- [x] Unset auth cookie if expired from server.
- [x] Auth check on pages.
- [x] auth on all menu objects
- [x] Move pageId back to context
- [x] Pages auth config extra checks
- [x] Filter menus


packages/engine/src/getContext.js
```diff
@@ -57,33 +57,32 @@ const blockData = ({
...
   }
-  rootContext.contexts[contextId] = {
+  lowdefy.contexts[contextId] = {
     id: contextId,
     blockId: block.blockId,
     eventLog: [],
-    pageId,
     requests: {},
-    root: rootContext,
+    root: lowdefy,
```
SamTolmay, line 77: Maybe we can rename root here to lowdefy as well?

Gervwyk: rename all the things 🐙

SamTolmay (approved):

PR #513, opened 2021-03-24
chore: add `Apache`

susiwen8:
Hi, my name is Sven and I'm a member of Apache ECharts PMC. Very glad to see `lowdefy` has chosen Apache ECharts. I came across and notice some issue in doc. It should be `Apache ECharts` rather than `ECharts` because right now it is TLP of Apache Foundation.   Could you merge this PR, so there would be no brand issue. Thanks, and excellent job you guys have done here,

### DIFF
--- a/packages/blocks/blocksECharts/README.md
+++ b/packages/blocks/blocksECharts/README.md
@@ -1,11 +1,11 @@
 # Lowdefy ECharts Blocks
 
-Lowdefy blocks for [ECharts](https://echarts.apache.org/), is a feature rich javascript charting library.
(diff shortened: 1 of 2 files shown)


Gervwyk:
Hi @susiwen8 . Thanks for advising!

Would it be preferred that we rename the block type `ApacheEChart` instead of `EChart` or is it fine to just correct the references in the docs?

susiwen8:
> Would it be preferred that we rename the block type ApacheEChart instead of EChart or is it fine to just correct the references in the docs?

@Gervwyk Just correct the references in the docs, no need to rename block type.

Gervwyk (approved):

PR #531, opened 2021-04-01
 Add a JavaScript operator.

Gervwyk:


packages/build/src/build/buildPages.js
```diff
@@ -30,8 +30,45 @@ Blocks:
...
+    return value;
+  };
+  const { requests, ...webBlock } = block;
+  webBlock.areas = JSON.parse(JSON.stringify(webBlock.areas || {}), stripContext);
+  const operators = new Set();
+  const pushOperators = (_, value) => {
+    if (type.isObject(value) && Object.keys(value).length === 1) {
+      const key = Object.keys(value)[0];
+      const [op, _] = key.split('.');
+      const operator = op.replace(/^(\_+)/gm, '_');
```
SamTolmay, line 50: maybe just `if (op.startsWith('_'))`?
Why do we need to regex replace here?

Gervwyk: Because it replaces all multiples of '_'  from the start with a single '_'.

packages/docs/operators/_ref.yaml
```diff
@@ -58,19 +58,51 @@ _ref:
       module.exports = transformer;
       ```
 
+      ###### Evaluating JavaScript
+
+      JavaScript can be evaluated during build time by referencing a JavaScript file to the `eval` argument. The result of `_ref` operator will be the exported result of the evaluated JavaScript using [`eval`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval). If the exported result is a function, `function.toString()` will be returned, useful for loading `code` for the [`_js`](/_js) operator. The `eval` argument should be the file path (relative to the root of the project) to a JavaScript file and exports the returned result.
```
SamTolmay, line 63: that exports...

Gervwyk:
Build will only pass if new cli is published.
```
Error: Invalid _ref definition {"_ref":{"eval":"templates/blocks/filterDefaultValue.js"}} in file templates/blocks/template.yaml.njk
```

SamTolmay (approved):

SamTolmay (approved):

PR #535, opened 2021-04-07
Add Knex connection

SamTolmay:


packages/docs/connections/Knex.yaml
```diff
@@ -0,0 +1,159 @@
...
+            ```yaml
+            connections:
+              - id: mysql
+                type: Knex
+                properties:
+                  client: postgres
+                  connection:
+                    host: '127.0.0.1'
+                    user : my_database_user
+                    password : my_database_password
```
Gervwyk, line 66: Maybe rather always use _secret even in these examples?

packages/docs/connections/Knex.yaml
```diff
@@ -0,0 +1,159 @@
...
+            - `tableName: string | object`: The name of the table to query from.
+
+            #### Examples
+
+            Build a query:
+            ```yaml
+            id: knexBuilder
+            type: KnexBuilder
+            connectionId: knex
+            properties:
```
Gervwyk, line 100: Add a tableName parameter?

packages/docs/connections/Knex.yaml
```diff
@@ -0,0 +1,159 @@
...
+                    - users
+                - where:
+                    - name
+                    - _state: name
+            ```
+
+            ### KnexRaw
+
+            #### Properties
+            - `query: string`: __Required__ - SQL query string.
```
Gervwyk, line 114: Not seeing tableName in the connection, so I guess all these should have a tableName parameter? Also in the examples.

packages/graphql/src/connections/Knex/KnexRaw/KnexRawSchema.json
```diff
@@ -0,0 +1,28 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Lowdefy Request Schema - KnexRaw",
+  "type": "object",
+  "required": ["query"],
+  "properties": {
+    "query": {
+      "type": "string",
+      "description": "SQL query string.",
+      "errorMessage": {
+        "type": "KnexRaw request property \"query\" should be a string."
```
Gervwyk, line 11: Again not sure if table name should be here, or must you say `FROM "tablename"` ... ?

Gervwyk (dismissed):

packages/docs/connections/Knex.yaml
Gervwyk: Just resolve if it is meant to be in the query like `FROM ....`.

Gervwyk (dismissed):

Gervwyk (dismissed):

Gervwyk (approved):

Gervwyk (approved):

PR #549, opened 2021-04-24
Prep operator args to handle void instance.

Gervwyk:
…


packages/operators/src/common/object.js
```diff
@@ -16,16 +16,43 @@
...
+  return args;
+};
 
 const metaInstance = {
-  hasOwnProperty: { namedArgs: ['on', 'prop'], validTypes: ['array', 'object'] },
+  hasOwnProperty: { namedArgs: ['on', 'prop'], validTypes: ['array', 'object'], prep },
 };
 
 const metaClass = {
   keys: { singleArg: true, validTypes: ['object'] },
```
SamTolmay, line 48: Do we want to use the prep function for keys, values and assign?

SamTolmay (approved):

PR #577, opened 2021-05-16
Custom javascript operators and actions, as well as header and body html.

Gervwyk:


packages/servers/serverDev/src/shell/public/modules/filterDefaultValue.js
```diff
@@ -0,0 +1,48 @@
+/*
```
SamTolmay, line 1: Remove this file

packages/servers/serverNetlify/src/shell/index.html
```diff
@@ -20,15 +20,39 @@
       <meta name="viewport" content="width=device-width, initial-scale=1" />
```
SamTolmay, line 20: Remove head.html and body.html

packages/docs/actions/JsAction.yaml
```diff
@@ -0,0 +1,220 @@
...
+      ```
+      (params: {
+        name: string,
+        args?: any[]
+      }): void
+      ```
+    description: |
+      The `JsAction` action is used to call a custom JavaScript function which was loaded onto the page using the `window.lowdefy.registerJsAction()`. This JavaScript function can be asynchronous. See [Custom JavaScript](/custom-javascript) for more details on how to register a new JavaScript action.
+
+      The returned result of the JavaScript function is accessible through the [`_actions`](/_actions) operator for subsequent action in the event action chain.
```
SamTolmay, line 31: subsequent actions

Gervwyk:
Happy with the changes. We can merge the PR. I'm blocked to approve since it is my PR.

SamTolmay (approved):

SamTolmay:
Closes #585

PR #673, opened 2021-06-29
Fix parse action error message

Gervwyk:


packages/engine/src/Actions.js
```diff
@@ -132,12 +131,24 @@ class Actions {
         params: parsedAction.params,
       });
     } catch (error) {
+      responses[action.id] = { error, index, type: action.type };
```
SamTolmay, line 134: At line 43: `responses[action.id] = error;`

packages/engine/src/Actions.js
```diff
@@ -132,12 +131,24 @@ class Actions {
...
+        location: block.blockId,
+      });
+      if (parserErrors.length > 0) {
+        // this condition is very unlikely since parser errors usually occur in the first parse.
+        throw { error: parserErrors[0], type: action.type, index };
+      }
       closeLoading();
       this.displayMessage({
-        defaultMessage: error.lowdefyMessage || 'Action unsuccessful',
+        defaultMessage: error.message,
```
SamTolmay, line 148: Do we not want to rather show more generic message to users?

packages/engine/src/Actions.js
```diff
@@ -132,12 +131,24 @@ class Actions {
         params: parsedAction.params,
       });
     } catch (error) {
+      responses[action.id] = { error, index, type: action.type };
+      const { output: parsedMessages, errors: parserErrors } = this.context.parser.parse({
+        actions: responses,
+        event,
+        arrayIndices,
+        input: action.messages,
```
SamTolmay, line 139: Only really need to parse `action.messages.error`

packages/engine/src/Actions.js
Gervwyk: So the actual error message by default a more standard thing to do, we have `messages.error` if the developer would like to make the response customized. Also, if a user told you the error was .... the chances of finding / fixing it are better. Thoughts?

packages/engine/src/Actions.js
Gervwyk: If we change this then test `operators are evaluated in params, skip and messages` fails. This means the `messages` field is not parsed entirely, but only a subset - the error field. As is, you could get the messages object with the schema fields.

SamTolmay (approved):

PR #679, opened 2021-07-01
Feature: Add Elasticsearch connection

Radiergummi:
Closes: --

As discussed in Discord, I implemented initial support for Elasticsearch as a connection in Lowdefy. This PR only includes support for searching the index: While Elasticsearch supports full CRUD for documents, it is not fully clear what is out of scope for Lowdefy, how these endpoints will be implemented, or what kind of abstraction may be put in place.  

The connection uses the official, supported, documented, and well maintained [Elasticsearch client by Elastic Co](https://github.com/elastic/elasticsearch-js). It supports all options exposed by the client, which is quite a lot; while the schema contains virtually _all_ available options, the documentation page focuses on the most used options and includes a link to the full reference. This is still lots of text, but I guess it helps that way. Let me know what you think!

 - A new connection has been added to the resolver. This should not have any implications on existing instances.
 - A new direct and two indirect dependencies have been added via `@elastic/elasticsearch`.

## Checklist

- [x] Tests added


packages/graphql/src/connections/Elasticsearch/ElasticsearchSearch/ElasticsearchSearch.js
```diff
@@ -0,0 +1,101 @@
...
+
+    const response = await client.search( {
+        ...params,
+        index: request.index || connection.index,
+        body,
+    } );
+
+    // Use the actual documents from the response as the base return value, then
+    // augment it with the additional, potentially interesting props.
+    const hits = response.body?.hits?.hits || [];
```
SamTolmay, line 49: @Gervwyk, I am not sure if our current build setup with babel targeting node 12 supports optional chaining. We have used the `get` function from `@lowdefy/helpers` for this in the past, but it might be better switch to optional chaining.

SamTolmay:
Tests failing because of missing codeclimate id secret in pull requests from an forked repo. Tests pass though, so this in not an issue. Maybe we should see if we could setup Github actions to not fail if that step fails.

packages/graphql/src/connections/Elasticsearch/ElasticsearchSearch/ElasticsearchSearch.js
Gervwyk: We've been wanting to switch get for optional chaining for a while. Think it would make sense to switch everything at once to keep the code base consistent. Should we add it as a chore for v4?

Gervwyk: Yes, we've been wanting to switch everything to optional chaining for a while. Maybe it is best to do it all at once for consistency in the codebase. Let's add it at a chore for v4?

Gervwyk:
Looks like the linter settings did not kick in for your IDE, but I'm happy to merge at is and fix as a chore before release.

@SamTolmay I guess learna will not know how to interpret the commit messages for release notes. Maybe we should just add the commit message instructions to the contribution notes.

packages/docs/connections/Elasticsearch.yaml
```diff
@@ -0,0 +1,156 @@
...
+              - ElasticsearchSearch
+
+            ### ElasticsearchSearch
+
+            The `ElasticsearchSearch` request searches the Elasticsearch index.
+
+            #### Properties
+            - `query: object`: The Elasticsearch query, expressed in the [JSON Query DSL](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html). If you need to pass custom request body attributes, you should wrap the query within another `query` property (in that case, the top-level `query` object will be used as `body`).
+            - `aggs: object[]`: The [aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html) to request from the server.
+            - `_source: string|string[]|boolean`: True or false to return the _source field or not, or a list of fields to return.
```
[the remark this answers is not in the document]
Radiergummi: That's a tough one, didn't think about that yet. I guess it's easier to rename the request properties to `source`, `source_include`, and `source_exclude`, then map them into the Elasticsearch client parameters (inside the resolver all operators should have been resolved already, right?).  
It's pretty obvious for users, I hope.

On the other hand, Elasticsearch has several properties in the responses that are also prefixed with an underscore (eg. `_id`, `_score`, `_source`, etc.), don't know about that..

packages/graphql/src/connections/Elasticsearch/ElasticsearchSearch/ElasticsearchSearch.js
Radiergummi: Ah, sorry! I'm so used to optional chaining by now, didn't think about support. I can swap them out for the helper if that's easier for now?

Radiergummi:
> Looks like the linter settings did not kick in for your IDE, but I'm happy to merge as is and fix as a chore before release.

Sorry, just ran prettier. Should fit in better now 🙂

packages/docs/connections/Elasticsearch.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+            #### Examples
+
+            ###### Authenticated connection:
+            ```yaml
+            connections:
+              - id: elasticsearch
+                type: Elasticsearch
+                properties:
+                  node: ELASTICSEARCH_HOST
```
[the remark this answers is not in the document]
Radiergummi: Fixed in f5ad31a

packages/graphql/src/connections/Elasticsearch/ElasticsearchSearch/ElasticsearchSearch.js
Gervwyk: Optional chaining is great. From a consistency point of view, let's use [get](https://github.com/lowdefy/lowdefy/blob/main/packages/helpers/src/get.js), you'll see we use it everywhere, then when we do #680 we'll convert it everywhere.

packages/docs/connections/Elasticsearch.yaml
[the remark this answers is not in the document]
Gervwyk: Just remembered that this is less of an issue than I thought. So we have 2 rules with operators:
1) starts with _
2) the operator object must only have 1 key

Thus for implementations like these the second rule will mostly cause the operator not to evaluate for that object. So there is a much smaller chance of name clashes here. Not really sure how / where to document this.
For now at least I think we can keep this as is, right @SamTolmay ?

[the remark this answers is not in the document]
SamTolmay: I think it would be best to keep it as is, there should be at least another field like query specified, for the request to work. So this happening should be an edge case. We can also add a warning to the docs if we want.

packages/graphql/src/connections/Elasticsearch/ElasticsearchSearch/ElasticsearchSearch.js
Radiergummi: Fixed in bba533a

packages/docs/connections/Elasticsearch.yaml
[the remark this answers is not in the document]
Radiergummi: Yeah, `_source` on its own doesn't produce a valid query, so that'd probably be a bug anyway.

Gervwyk (approved):

SamTolmay:
Hi @Radiergummi 

Sorry I could only get to this PR now.

I have changed both the request schema as well as the response schema.

In terms of the request schema, I have moved the `query` and `aggs` properties back to the `body` property, so it matches with the elasticsearch client request schema. We have found that matching the underlying library helps with future maintainability and documentation. If we change that schema, we need to document it well, and users can't use other examples.

In terms of the response schema, the schema you had was not JSON serializable. So I removed the `Infinity` in the total count. I also reworked the way the hits are transformed.

Please have a look, and see if you are happy with this, and let me know if you have any suggestions.

Radiergummi:
Hi @SamTolmay,

LGTM! The changes make sense; I guess I was just too deep in. I've just started working on our first internal application relying on Lowdefy and Elasticsearch, so I can follow up with new features soon, I hope.

SamTolmay:
Hi @Radiergummi.

That's great! You always find out how the connection should work and what features you need once you build a few real-world apps using it. Thanks again for the contribution.

Issue #631, opened 2021-06-14
Change case operator


SamTolmay:
https://github.com/lowdefy/lowdefy/discussions/540#discussioncomment-614082

PR #702, opened 2021-07-22
Change case operator

JohannMoller:
Closes: #631

This pull request adds a new `_change_case` operator using the [change-case](https://www.npmjs.com/package/change-case) package to transform a string between cases using a list of methods.

## Checklist

- [x] Tests added


packages/operators/src/common/change_case.js
```diff
@@ -0,0 +1,92 @@
...
+
+const convertObject = ({ methodName, on, options = {} }) => {
+  const result = {};
+  const keyConverter = options.convertKeys
+    ? (key) => changeCase[methodName](key, options)
+    : (key) => key;
+  const valueConverter = get(options, 'convertValues', { default: true })
+    ? (val) => changeCase[methodName](val, options)
+    : (val) => val;
+
```
SamTolmay, line 52: Maybe rather do `([key, val]) =>` instead

packages/operators/test/common/change_case.test.js
```diff
@@ -0,0 +1,504 @@
...
+          convertKeys: false,
+        },
+      },
+      location: 'locationId',
+      methodName: 'capitalCase',
+    })
+  ).toEqual({ field_1: 'Test String 1', field_2: 'Test String 2' });
+});
+
+test('_change_case.capitalCase on: object, options: {}', () => {
```
SamTolmay, line 98: Add that convertValue is default true in test name

Gervwyk (approved):

Gervwyk (approved):

JohannMoller:
I didn't add examples of change-case's own options, only our `convertKeys` and `convertValues` options. Is that okay?

packages/docs/operators/_change_case.yaml
```diff
@@ -0,0 +1,287 @@
...
+
+_ref:
+  path: templates/operators.yaml.njk
+  transformer: templates/operatorsMethodTransformer.js
+  vars:
+    pageId: _change_case
+    pageTitle: _change_case
+    filePath: operators/_change_case.yaml
+    description: |
+      The `_change_case` operator uses the [`change-case`](https://www.npmjs.com/package/change-case) package to transform an input between camelCase, PascalCase, Capital Case, snake_case, param-case, CONSTANT_CASE and others.
```
Gervwyk, line 23: Please add a list of all possible cases instead of etc.

packages/docs/operators/_change_case.yaml
```diff
@@ -0,0 +1,287 @@
...
+    arguments: |
+      The `_change_case` operator takes `on` and `options` as arguments.
+      The `on` argument is the input to be transformed and can be a `string`,  `array` or `object`.
+      The `options` argument is an object with the following properties:
+        - `splitRegexp`: The regular expression used to split `on` into word segments.
```
Gervwyk, line 28: Can we add tests for these regex options?

When passing in reg expressions to operators you need to test for the option and create a Regex from the input string to pass to the lib. 

Have a look at the implementation for `_regex`.

Gervwyk (changes requested):

packages/docs/operators/_change_case.yaml
```diff
@@ -0,0 +1,343 @@
...
+        on:
+          foo: 'bar'
+      ```
+      Returns: `{ "foo": "Bar" }`
+
+      ###### Object with options.convertKeys: true:
+      ```yaml
+      _change_case.capitalCase:
+        on:
+          foo: 'bar'
```
SamTolmay, line 74: Add option here

packages/docs/operators/_change_case.yaml
```diff
@@ -0,0 +1,343 @@
...
+      ```
+      Returns: `{ "Foo": "Bar" }`
+
+      ###### Object with options.convertValues: false:
+      ```yaml
+      _change_case.capitalCase:
+        on:
+          foo: 'bar'
```
SamTolmay, line 82: Same as above

Gervwyk (approved):

Issue #727, opened 2021-08-10
Elasticsearch index, update and delete requests.


SamTolmay:
The following addition requests types are required to support full CRUD operations on Elasticsearch:

- `ElasticsearchIndex `
- `ElasticsearchUpdate`
- `ElasticsearchUpdateByQuery `
- `ElasticsearchDelete`
- `ElasticsearchDeleteByQuery `

Thes should mirror the corresponding client/API methods as much as possible.

PR #730, opened 2021-08-11
Feature/add elasticsearch write queries

Radiergummi:
Closes: #727

This PR adds support for creating, updating and deleting documents stored in Elasticsearch. It also includes actions to delete and update documents by query, effectively providing bulk operation support.

## Checklist
- [x] Tests added


packages/docs/connections/Elasticsearch.yaml
```diff
@@ -88,6 +90,11 @@ _ref:
 
             Request types:
               - ElasticsearchSearch
+              - ElasticsearchIndex
```
SamTolmay, line 93: Can you please sort requests in alphabetic order? And do the same for the order below

SamTolmay (changes requested):
Thanks, this is looking great. I like the examples, they are nice and clear. Can you please just sort the requests in alphabetical order to be consistent with the other request docs?

packages/docs/connections/Elasticsearch.yaml
Radiergummi: ✅

SamTolmay (approved):

Gervwyk (approved):

PR #742, opened 2021-08-12
Throw action

JohannMoller:
Closes: #731 

This pull request adds the Throw action which will throw an error, log it to the console and break the action chain.

## Checklist

- [x] Tests added


packages/docs/actions/Throw.yaml
```diff
@@ -0,0 +1,63 @@
...
+      ```
+      (params: {
+        throw?: boolean,
+        message?: string,
+        metaData?: any
+      }): void
+      ```
+    description: |
+      The `Throw` action is used throw an error to the user and log to the console. If `throw: true`, the `Throw`
+      action will throw an error, and this will stop the execution of actions that are defined after it. If the actions is not thrown, the `Throw` action will do nothing and the actions defined after it will be executed.
```
SamTolmay, line 31: Maybe change "If the actions is not thrown," to "If the action does not throw,"

packages/docs/actions/Throw.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    params: |
+      - `throw: boolean`: Throws an error and stops the action chain when `true` or continues the action chain when `false` or undefined.
+      - `message: string`: The error message to show to the user and log to the console if `throw: true`.
+      - `metaData: any`: Data to log to the console if `throw: true`.
+
+    examples: |
```
SamTolmay, line 38: Please add an example of throwing silently by setting `messages.error to false`.

Gervwyk (approved):

PR #749, opened 2021-08-13
Actions async option

### LINKED ISSUE #745: Add async options to actions
The action chain will continue without awaiting the action.

JohannMoller:
Closes #745 

The pull request adds an `async` option to actions which will allow them to be executed asynchronously in an event's action chain. 

Note: if this action throws an error, the event chain still completes successfully.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/engine/src/Actions.js
+++ b/packages/engine/src/Actions.js
@@ -27,18 +27,46 @@ class Actions {
     this.actions = actions;
   }
 
+  async callAsyncAction({ action, arrayIndices, block, event, index, responses }) {
+    try {
+      const response = await this.callAction({
+        action,
+        arrayIndices,
+        block,
+        event,
+        index,
+        responses,
+      });
+      responses[action.id] = response;
+    } catch (error) {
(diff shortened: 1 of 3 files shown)


SamTolmay (approved):

packages/engine/src/Actions.js
```diff
@@ -27,18 +27,46 @@ class Actions {
...
-        responses[action.id] = response;
+        if (action.async === true) {
+          this.callAsyncAction({
+            action,
+            arrayIndices,
+            block,
+            event,
+            index,
+            responses,
+          });
```
Gervwyk, line 58: @SamTolmay Should we not set
```js
this.callAsyncAction({
            action,
            arrayIndices,
            block,
            event,
            index,
            responses,
          }).then( (response) => { responses[action.id] = response; });
```
we need to log the action to the responses object.

Gervwyk (changes requested):

Gervwyk (approved):

Issue #737, opened 2021-08-12
Validate to apply pattern match on params.


Gervwyk:
Only raise validation errors on blocks that match validation of block id.

PR #756, opened 2021-08-16
Validate on regex match and move showValidation to blocks.

Gervwyk:
Closes: #737 

The `Validate` action can now be used to partially validate `state`. This enables the validation of multiple forms per page. `context.showValidationErrors` has also been removed and replaced by `block.showValidation` on a block level.  `showValidation` flags are only raised after Validation has been called on specified blocks.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/engine/src/Blocks.js
+++ b/packages/engine/src/Blocks.js
@@ -217,6 +217,7 @@ class Blocks {
     const initState = serializer.copy(initWithState || this.context.state);
     this.loopBlocks((block) => {
       block.update = true;
+      block.showValidation = false;
       if (get(block, 'meta.category') === 'input' || get(block, 'meta.category') === 'list') {
         let blockValue = get(initState, block.field);
         if (type.isUndefined(blockValue)) {
@@ -368,13 +369,13 @@ class Blocks {
             }
           }
         });
-        if (this.context.showValidationErrors && validation.length > 0) {
+        if (validation.length > 0) {
           block.validationEval.output.status = 'success';
         }
         if (validationWarning) {
           block.validationEval.output.status = 'warning';
         }
-        if (this.context.showValidationErrors && validationError) {
+        if (validationError) {
           block.validationEval.output.status = 'error';
         }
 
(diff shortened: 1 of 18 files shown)


packages/engine/test/State.test.js
```diff
@@ -18,34 +18,28 @@ import State from '../src/State';
...
   const context = {
-    showValidationErrors: false,
     state: {},
   };
   const state = new State(context);
   state.set('a', 1);
   expect(context.state).toEqual({ a: 1 });
 });
 
-test('resetState and clear showValidationErrors flag', () => {
```
SamTolmay, line 29: We need to remember to test that `block.showValidation` gets set to false  when we reset

packages/docs/actions/Validate.yaml
```diff
@@ -23,46 +23,83 @@ _ref:
...
       The `Validate` action is used to validate a users input, usually before information is inserted into a database using a request.
       It is used in conjunction with the `required` and `validate` fields on input blocks. If the validation fails, the `Validate`
       action will fail, and this will stop the execution of actions that are defined after it.
 
-      The first time a `Validate` action is called, validation error are shown to the user. The `Reset` action resets this.
+      The first time a `Validate` action is called, validation errors and warnings are shown to the user. The `Reset` action resets the validation status and the page `state`. The `ResetValidation` action resets only the validation status.
 
-      Validate only validates the input blocks that are inside the context it is called from. Therefore, if you use two contexts
-      to separate two forms, only the one form will be validated.
+      Validate only validates the blocks that are on the page which matches the blockIds matching the `Validate` params criteria.
```
SamTolmay, line 39: Maybe rather something like:
```
If the `Validate` action `blockIds` or `regex` params are used to limit which blocks are validated, only the matched blocks are validated, and validation results are shown for only those matched blocks.
```

packages/docs/actions/Validate.yaml
```diff
@@ -23,46 +23,83 @@ _ref:
...
-      - id: validate_one
+      - id: validate_foo
         type: Validate
         params: my_text_input
       ```
 
       ###### Validate a list of inputs:
       ```yaml
-      - id: validate_one
+      - id: validate_foo
```
SamTolmay, line 71: Maybe we should rather do `validate_all`, `validate_my_input`, `validate_my_inputs` for these ids?

packages/engine/src/Blocks.js
```diff
@@ -562,7 +573,10 @@ class Blocks {
           required: block.requiredEval.output,
           layout: block.layoutEval.output,
           style: block.styleEval.output,
-          validation: block.validationEval.output,
+          validation: {
+            ...(block.validationEval.output || {}),
+            status: block.showValidation ? (block.validationEval.output || {}).status : null,
```
SamTolmay, line 578: Do we have tests for this check?

SamTolmay (approved):

PR #762, opened 2021-08-19
Debounce events

### LINKED ISSUE #724: Add event debouncing
Sometimes while handling user input, we face the problem of input being provided _too fast_, for example while tracking mouse movement, scrolling speed or keyboard presses. In these cases, the browser fires events rapidly, sometimes faster than the event listener finishes: The work performed is wasted, because the result is outdated before the callback even finishes.  
We can solve this issue by [debouncing](https://davidwalsh.name/javascript-debounce-function): A debounced callback will only be executed if the event we listen to didn't fire for a given amount of time. This might happen if the user stops scrolling, or pauses pressing keys on their keyboard.

It would be great to have the option to debounce events fired in Lowdefy applications. As a concrete example, the [`AutoComplete` block](https://docs.lowdefy.com/AutoComplete) emits the `onChange` event every time the input value changes. If we supply this block with suggestions fetched from a remote server, a separate request will be fired for every character entered in quick succession, leading to a lot of (potentially expensive) requests that carry the entire application state with them (see #641).

@SamTolmay suggested the following schema for debounced events:
```yaml
events:
  onClick:
    debounce:
      delay: 300
      trailingEdge: true
    try:
      - id: request
        type: Request
        # ...
    catch: []
```

This [freeCodeCamp article](https://www.freecodecamp.org/news/javascript-debounce-example/) provides a good introduction into debouncing.

Gervwyk:
Closes #724 

Allow for both leading and trailing edge event debouncing settings. See #724 

## Checklist

- [x] Tests added


packages/engine/src/Events.js
```diff
@@ -52,37 +55,90 @@ class Events {
...
+    };
 
-    eventDescription.history.unshift(result);
-    this.context.eventLog.unshift(result);
-    eventDescription.loading = false;
-    this.block.update = true;
-    this.context.update();
+    // no debounce
+    if (type.isNone(eventDescription.debounce)) {
+      return await actionHandle();
```
SamTolmay, line 95: No need for await here

SamTolmay (approved):

PR #763, opened 2021-08-19
Reference resolver functions

### LINKED ISSUE #753: Add a resolver function to _ref operator
It would be useful if custom JS can be run when resolving a reference by the `_ref` operator. Some example use cases might be:
- To load a file from a different, "defaults" folder if the reference isn't found.
- To read the file from a Github repository instead of the local file system.
- To generate config instead of reading the file system at all.

It should be possible to set the resolver for all `_ref` operators, or to specify a resolver function for just one reference.

SamTolmay:
Closes #753 

This PR adds a new feature to `_ref` operators, called resolvers. Resolvers are user-defined functions that overwrite the default way configuration files are read from the file system.

## Checklist

- [x] Tests added


packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -199,6 +199,8 @@ _ref:
 
             Templating is used extensively to create the Lowdefy docs (these docs are a Lowdefy app). You can look at how they are used [here](https://github.com/lowdefy/lowdefy/tree/main/packages/docs).
 
+            The `_ref` operator can also be extended with custom JavaScript functions. A `resolver` function can be specified, to which can overwrite the default way configuration files are read from the filesystem. A `transformer` function can be used to transform the value returned by the `_ref` operator.
```
JohannMoller, line 202: "...to which..." typo

packages/docs/operators/_ref.yaml
```diff
@@ -161,3 +163,62 @@ _ref:
...
+      const fs = require('fs')
+      const path = require('path')
+      const { promisify } = require('util');
+
+      const readFilePromise = promisify(fs.readFile);
+
+      async function useLocalOrSharedConfig(path, vars, context) {
+        const localFile =  await readFilePromise(path.resolve(path), 'utf8');
+
+        if (localFile) return localFile
```
Gervwyk, line 182: Just add `;` for both return lines.

Gervwyk (approved):

Gervwyk (approved):

PR #772, opened 2021-08-21
Fix card title

SamTolmay:

This pull request fixes a bug where the `Card` block title area was not rendering properly due to the HTML support introduced in v3.20.0.

It also fixes user-defined block types, that were not being used.

## Checklist

- [x] Tests added


Gervwyk (dismissed):
We need to check if the same issue for Card does not apply to other blocks.

Gervwyk (approved):

PR #791, opened 2021-08-24
onMount and onMountAsync events on all blocks.

### LINKED ISSUE #778: onEnterAsync and onInitAsync on all blocks
#### Describe the feature

`onEnter` and `onInit` are very specific page events where the order / timing of when events are triggered matters. However, for `onEnterAsync` and `onInitAsync` the timing of when these events are triggered does not matter, only that it is triggered.  

Thus `onEnter` and `onInit` should remain only on page blocks, but `onEnterAsync` and `onInitAsync` could perhaps be triggered on all blocks. This would help make template abstraction easier since the required onEnterAsync events can be contained on the template and do not need to be promoted to the page block. Consider if we can / should batch these requests to the server.

Not sure if this is a better option than adding react lifecycle events to blocks, like onWillMount, onDidMount and onUnmount.

Gervwyk:
Closes #778 

All blocks now has the `onMount` and `onMountAsync` events.

https://www.loom.com/share/25e8bbe4795641a4a46311daa57de6eb

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/renderer/src/page/DisplayMessage.js
+++ b/packages/renderer/src/page/DisplayMessage.js
@@ -34,20 +34,24 @@ const Block = ({ methods }) => {
               remoteEntryUrl: `https://blocks-cdn.lowdefy.com/v${packageJson.version}/blocks-antd/remoteEntry.js`,
             },
           }}
-          Loading={''}
-          render={(Comp) => (
-            <Comp
-              blockId="__display_message"
-              key="__display_message"
-              methods={{
-                makeCssClass,
-                registerMethod: methods.registerMethod,
-                triggerEvent: () => undefined,
-              }}
-              properties={{}}
-            />
-          )}
-        />
+        >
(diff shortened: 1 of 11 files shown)


packages/renderer/src/page/DisplayMessage.js
```diff
@@ -34,20 +34,24 @@ const Block = ({ methods }) => {
               remoteEntryUrl: `https://blocks-cdn.lowdefy.com/v${packageJson.version}/blocks-antd/remoteEntry.js`,
             },
           }}
-          Loading={''}
```
SamTolmay, line 37: the component name should be renamed from `Block` to `DisplayMessage`

Gervwyk:
https://www.loom.com/share/c5ec41732200488f99392481e905a031

SamTolmay (approved):

PR #818, opened 2021-08-31
Default parameter of _var operator should allow values like empty string, zero or boolean false.

### LINKED ISSUE #819: Default parameter of _var operator replaces values that JS evaluates as boolean false (i.e. zero, empty string, boolean false) with nulls.
#### Describe the bug

Default parameter of _var operator replaces values that JS evaluates as boolean false (i.e. zero, empty string, boolean false) with nulls. It should only do that if provided value is not part of the JSON spec or value is undefined.

#### To Reproduce

YAML:
```
  key_empty_str:
    _var:
      name: var1
      default: ''
  key_false:
    _var:
      name: var2
      default: false
  key_zero:
    _var:
      name: var3
      default: 0`,
      },
```
Expected:
```
{
  key_empty_str: '',
  key_false: false,
  key_zero: 0
}
```

Result:
```
{
  key_empty_str: null,
  key_false: null,
  key_zero: null
}
```

#### Lowdefy Version

3.21.2

mat02:
Closes #819

Default parameter of _var operator should allow values like empty string, zero or boolean false. Default value should be set to null only if parameter is undefined, explicitly set to null or is not part of the JSON spec (like Inf or NaN).

## Checklist

- [x] Tests added


packages/build/src/build/buildRefs/buildRefs.test.js
```diff
@@ -396,22 +396,74 @@ describe('vars', () => {
...
+      name: var2
       default: default`,
       },
     ];
     mockReadConfigFile.mockImplementation(readConfigFileMockImplementation(files));
     const res = await buildRefs({ context });
     expect(res).toEqual({
       ref: {
-        key: 'default',
+        key1: null,
```
SamTolmay, line 418: @Gervwyk, @mat02 

I am not sure about this behaviour. If the var is defined, and has value `null`, the default value is not used. I guess this is the behaviour of the `get` function, but this might cause unexpected bugs, since, for example, the `_ref` operator returns `null` if the file is not found.

SamTolmay:
Thanks for the PR @mat02. We are very happy to have you as a contributor to the project 😀

packages/build/src/build/buildRefs/buildRefs.test.js
mat02: Hmmm, I didn't think it through. Maybe defining var and then setting its value to null is not something that is really needed. Then, `type.isUndefined` should be used as a test function (as you suggested on Discord), not `test.isNone`. Should I change it

SamTolmay: The issue is that the value of the var is `null`, not the value of the default value. Changing to `type.isUndefined` shouldn't change anything (I believe).

The issue is with the Lowdefy `get` helper. The get function tries to get a value and receives `null`, which is a valid value, and therefore doesn't return the default. All getters in Lowdefy like `_state` will have this behaviour too.

SamTolmay: I am going to merge this since this case we are discussing is a separate issue with the get helper.

SamTolmay (approved):

PR #931, opened 2021-11-03
Add init page

### LINKED ISSUE #815: App onLoad and onLoadAsync events
#### Describe the feature

An app level init event which triggers on app load for any page. Perhaps consider a onPageLoad events also which triggers when any page is loaded for the first time.

mat02:
Closes #815 

This PR loads selected page when app is started, then redirects to the page specified in the url. This allows to define actions that will be executed when app is first started and/or before each page is loaded in a single place as onInit/onInitAsync or onEnter/onEnterAsync events. The usecase for that could be fetching global data, that is later used on multiple pages or reseting inactivity timer when new page is loaded.

## Checklist

- [x] Tests added


mat02:
Upss, I will fix the broken tests and upload revised version.

SamTolmay (approved):

PR #985, opened 2021-11-23
Generate block ids during build

### LINKED ISSUE #920: Generate the unique id for blocks that we use in engine during build

SamTolmay:
Closes #920 

## Checklist

- [ ] Tests added


Gervwyk (approved):

mingfang:
I think we can go one step further with this and make block id's optional/anonymous if our application will never reference them.  

For example, there's really no value in setting the block id here.  The build process should just assign these blocks with random ids.
https://github.com/lowdefy/lowdefy/pull/985/files#diff-52eacedece3a2171e1cea6b50cffb14df7776e1674173f9a5b24f461c7c87dd7R855

PR #1053, opened 2022-01-13
V4 redis connection plugin

correiarmjoao:

Add connection-redis plugin and documentation.

## Checklist

- [x] Tests added


Gervwyk (dismissed):

packages/plugins/connections/connection-redis/package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "contributors": [
+    {
+      "name": "Sam Tolmay",
+      "url": "https://github.com/SamTolmay"
+    },
+    {
+      "name": "Gerrie van Wyk",
+      "url": "https://github.com/Gervwyk"
+    }
```
SamTolmay, line 22: @correiarmjoao, please add yourself as a contributor here

packages/plugins/connections/connection-redis/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "scripts": {
+    "build": "yarn swc",
+    "clean": "rm -rf dist",
+    "prepare": "yarn build",
+    "swc": "swc src --out-dir dist --config-file ../../../../.swcrc --delete-dir-on-start --copy-files",
+    "test": "jest --coverage"
+  },
+  "dependencies": {
+    "@lowdefy/helpers": "4.0.0-alpha.5",
+    "redis": "^4.0.1"
```
SamTolmay, line 45: Please pin this dependency to a fixed version. We use pinned dependencies across the project. 
```
"redis": "4.0.1"
```

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.js
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error(`Connection refused.`);
+  }
+
+  if (!type.isFunction(client[command.toUpperCase()])) {
+    throw new Error(`Invalid redis command "${command}".`);
+  }
+
+  if (!type.isArray(parameters)) {
+    throw new Error(
+      `Invalid command, command "${command}" parameters should be an array, received ${JSON.stringify(
```
SamTolmay, line 45: Maybe this will be more clear if the message reads:
```
Invalid parameters, command "${command}" parameters should be an array,
```

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.js
```diff
@@ -0,0 +1,74 @@
...
+  }, {});
+
+  try {
+    const commandReturn = await client[command.toUpperCase()](...parameters, upperCaseModifiers);
+    await client.quit();
+    return commandReturn;
+  } catch (error) {
+    client.quit();
+    throw new Error(
+      `Invalid command "${command}" parameters, received ${JSON.stringify(parameters)}.`
```
SamTolmay, line 63: @Gervwyk In other requests we throw the original error to make it easier for developers to debug:
```js
try {
  // ...
} catch (error) {
  client.quit();
  throw error
}
```
I think we should do this here?

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.test.js
```diff
@@ -0,0 +1,57 @@
...
+  );
+});
+
+test('checkRead should be false', async () => {
+  expect(checkRead).toBe(false);
+});
+
+test('checkWrite should be false', async () => {
+  expect(checkWrite).toBe(false);
+});
```
SamTolmay, line 57: Please add tests with a mocked redis connection to test that the client and command functions receive the correct arguments, all the commands are converted to uppercase, errors like invalid command are thrown correctly etc.

This is a possible example of the mocks should look like, but I have not been able to test/run this so I am not sure that it will work:
```js
jest.mock('redis', () => {
  return {
    createClient: mockCreateClient,
  };
});

const mockCreateClient = jest.fn();

let mockCreateClient;

beforeAll(async () => {
  mockCreateClient = (await import('redis')).createClient;
});

test("mocked client", () => {
  const client = {
    on: jest.fn(),
    connect: jest.fn(),
    GET: jest.fn(),
  }
  mockCreateClient.mockImplementation(() => client);
  expect(mockCreateClient.mock.calls).toEqual(/*...*/);
  expect(client.connect.mock.calls).toEqual(/*...*/);
  expect(client.GET.mock.calls).toEqual(/*...*/);
})
```

SamTolmay (changes requested):
Thanks @correiarmjoao, this is looking great! I am very happy with how nicely the library is integrated in a flexable way. Please have a look at my comments and add your feedback.

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.js
Gervwyk: Yes, we always pass the original error. This should change, thanks.

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.test.js
correiarmjoao: @SamTolmay The example works mock implementation works fine. 

Though I spent last Friday trying to make it work just to realise jest mocks were not working correctly.
It seems to be an issue with jest and swc, you can see it in [this thread](https://github.com/swc-project/jest/issues/14).

It was fixed with a patch [here](https://github.com/swc-project/jest/pull/44#issuecomment-980601069).

This issue is affecting the tests on the other plugins too, but for now I bumped the @swc/core in the connection-redis only and the mocks work correctly there.

SamTolmay: Okay thanks. I think we might have been running into this too. We will update the version for the other packages once we have merged this PR.

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.test.js
```diff
@@ -20,6 +20,161 @@ import Redis from './Redis.js';
...
+  };
+  const response = 'responseValue';
+  const res = await Redis({ request, connection });
+  expect(mockedCreateClient.mock.calls).toEqual([[connection.connection]]);
+  expect(client.connect).toHaveBeenCalledTimes(1);
+  expect(client.GET.mock.calls).toEqual([['key', { A: true, B: false }]]);
+  expect(client.quit).toHaveBeenCalledTimes(1);
+  expect(res).toEqual(response);
+});
+
```
SamTolmay, line 68: Please add a test for the case when the connection url is a string. (Line 22 in `Redis.js`)

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.test.js
```diff
@@ -20,6 +20,161 @@ import Redis from './Redis.js';
...
+  };
+  const request = {
+    command: 'get',
+    parameters: ['key'],
+    modifiers: {
+      a: true,
+      b: false,
+    },
+  };
+  await expect(Redis({ request, connection })).rejects.toThrow('Connection refused.');
```
SamTolmay, line 92: We should also throw the error returned by the redis client and not a generic error in this case. Sorry, I missed it in the previous review. Could you please update that and then update this test?

SamTolmay (changes requested):
This is looking great, the tests are nice and clear and easy to read. I have just requested a few small changes and then I am happy to merge 😀

SamTolmay:
@correiarmjoao Please also merge the `v4` branch into this branch as we require the branch to be up to date with the base branch before we can merge a PR, and we cannot do this since the edits from maintainers are not allowed.

packages/plugins/connections/connection-redis/src/connections/Redis/Redis/Redis.test.js
correiarmjoao: @SamTolmay 
No worries, I updated the test and the code. In this case we will not need the try/catch and client.on() will throw it on the callback.

correiarmjoao:
@SamTolmay
Made the changes and merged the v4 branch.
I found an issue regarding the "Allow edits from maintainers" now showing for me, it seems to be related with the fact that the merge request is being made from a repo that belongs to an organisation which I'm a member of, instead of a repo that is owned by me.

SamTolmay (approved):

Gervwyk (approved):

PR #1064, opened 2022-01-20
Public folder for next server.

Gervwyk:
Closes #ISSUE_NUMBER

Copies the content of the public folder into the next server public folder.

## Checklist

- [ ] Tests added


packages/utils/node-utils/src/cleanDirectory.js
```diff
@@ -14,10 +14,10 @@
   limitations under the License.
 */
 
-import rimraf from 'rimraf';
+import fsExtra from 'fs-extra';
 
 async function cleanDirectory(dirPath) {
-  await new Promise((resolve) => rimraf(dirPath, resolve));
+  await new Promise((resolve) => fsExtra.emptyDir(dirPath, resolve));
 }
```
SamTolmay, line 21: It looks like fs-extra returns a promise if no callback is supplied, so we should be able to do something like:
```js
async function cleanDirectory(dirPath) {
  await fsExtra.emptyDir(dirPath);
 }
```

SamTolmay (approved):

PR #1067, opened 2022-01-23
feat: Add CLI, dev and prod server port settings.

Gervwyk:

Add CLI, dev and prod server port settings.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/cli/src/commands/dev/runDevServer.js
+++ b/packages/cli/src/commands/dev/runDevServer.js
@@ -27,7 +27,7 @@ async function runDevServer({ context }) {
       env: {
         ...process.env,
         LOWDEFY_DIRECTORY_CONFIG: context.directories.config,
-        PORT: context.commandLineOptions.port,
+        PORT: context.options.port,
       },
     },
     silent: false,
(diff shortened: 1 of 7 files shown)


packages/server-dev/src/manager/run.mjs
```diff
@@ -28,7 +28,7 @@ async function run() {
     const serverPromise = startServer(context);
     await wait(800);
     // TODO: set correct port
```
SamTolmay, line 30: Remove TODO comment

packages/cli/src/commands/dev/runDevServer.js
```diff
@@ -26,7 +26,9 @@ async function runDevServer({ context }) {
       cwd: context.directories.devServer,
       env: {
         ...process.env,
+        LOWDEFY_PACKAGE_MANAGER: context.packageManager,
         LOWDEFY_DIRECTORY_CONFIG: context.directories.config,
+        PORT: context.options.port,
```
SamTolmay, line 31: Sjouldn't this rather be something like `LOWDEFY_PORT` or `LOWDEFY_SERVER_PORT`?

Gervwyk: Nextjs looks at PORT. https://nextjs.org/docs/api-reference/cli#development
So if we set PORT, then the Next server is configured. Think we should keep it the same?

SamTolmay (approved):

PR #1073, opened 2022-01-25
Connection Mongodb V4

correiarmjoao:

Fixed plugin structure to work with version 4.

## Checklist

- [ ] Tests added


correiarmjoao:
Mongodb

packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/getCollection.js
```diff
@@ -14,7 +14,8 @@
   limitations under the License.
 */
 
-import { MongoClient } from 'mongodb';
```
SamTolmay, line 17: Do we get an error if we import like this?

correiarmjoao: Yes we get a named export not found when imported in the build package.
`SyntaxError: Named export 'MongoClient' not found. The requested module 'mongodb' is a CommonJS module, which may not support all module.exports as named exports.`

packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/serialize.js
```diff
@@ -14,13 +14,15 @@
   limitations under the License.
 */
 
-import { ObjectID } from 'mongodb';
```
correiarmjoao, line 17: Same error here.
`SyntaxError: Named export 'ObjectID' not found. The requested module 'mongodb' is a CommonJS module, which may not support all module.exports as named exports.`
Also 'ObjectID' was flagged as deprecated.

packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/MongoDBAggregation/schema.js
```diff
@@ -0,0 +1,26 @@
+export default {
```
Gervwyk, line 1: Hi @correiarmjoao 
We need to add the license comment block to all the schema files. It's only the json files which do not include a license block.

Gervwyk (changes requested):
Hi @correiarmjoao 
Please add license comments to all new files. (schema.js)

packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/getCollection.js
correiarmjoao: @SamTolmay This will not be a problem in the new major version of MongoDB driver. It already uses named exports.

Gervwyk (dismissed):

SamTolmay (changes requested):
Please revert the formatting on the `.pnp.cjs` file.

SamTolmay (dismissed):

SamTolmay (approved):

PR #1076, opened 2022-01-25
Connection Google Sheets V4

correiarmjoao:

Fixed plugin structure to work with version 4.

## Checklist

- [ ] Tests added


correiarmjoao:
GoogleSheets

packages/plugins/connections/connection-google-sheets/src/connections/GoogleSheet/mingoAggregation.js
```diff
@@ -16,18 +16,18 @@
...
 
-useOperators(OperatorType.ACCUMULATOR, accumulatorOperators);
-useOperators(OperatorType.EXPRESSION, expressionOperators);
-useOperators(OperatorType.PIPELINE, pipelineOperators);
-useOperators(OperatorType.PROJECTION, queryOperators);
-useOperators(OperatorType.QUERY, projectionOperators);
+useOperators(OperatorType.ACCUMULATOR, accumulatorOperators.default || accumulatorOperators);
+useOperators(OperatorType.EXPRESSION, expressionOperators.default || expressionOperators);
+useOperators(OperatorType.PIPELINE, pipelineOperators.default || pipelineOperators);
+useOperators(OperatorType.QUERY, queryOperators.default || queryOperators);
```
correiarmjoao, line 29: Operator types were changed here.

Import * as '...' works well on the connection plugin, but when the package is being imported in the build package  to create the types this import is returning a different object structure.

Inside the connection package it returns:
{
 export1: function,
 export2: function,
 ...
}
When its being called inside the build package it returns:
{
  export1: function,
  export2: function,
  ...,
  __esModule: true,
  default: {
    export1: function,
    export2: function,
    ...
 }
}

I'm unsure about the way I "fixed it", have you guys had any similar problem with other plugins?

SamTolmay: Yes we have been having issues since ewe converted to ES Modules to simplify the project build and move to the new JS standard.

I think doing something like this is needed. Maybe you could do:
```js
useOperators(OperatorType.ACCUMULATOR, accumulatorOperators.__esModule ? accumulatorOperators : accumulatorOperators.default);
```

correiarmjoao: The problem with accumulatorOperators.__esModule is that it returns true for both imports, even though it's not an explicit key when importing inside the connection package. 
I could still use the ternary operator with default key to make it clear.

SamTolmay: In that case I think it is better like it is, and adding a comment explaining why you are doing this would be best for other developers looking at this in the future.

Gervwyk (approved):

SamTolmay (changes requested):
Please revert the formatting on the `.pnp.cjs` file.

SamTolmay (dismissed):

PR #1078, opened 2022-01-26
Dev server install plugins and rebuild next app.

### LINKED ISSUE #941: Create development next app

SamTolmay:
Closes #941 

This pull request adds a file watcher that watches for files that should trigger a Next build or npm install if they change. Some files, like plugin imports are written by lowdefy build, which triggers a file watch event even if the file content does not change, so a file content hash is used to keep track of changes to the files.

Previously, processes were started by calling `npm build`, `npm start`, etc. This introduces an additional process that spawns the actual process we want to run. This is less of a problem for short lived processes like build, but has issues for long lived processes like server. For example, the server process is not killed properly if the npm process is killed, or throws an error.

We need to do something similar for processes spawned by the CLI.

The `writeBuildArtifact` function in the build package was also updated to have a similar signature to the node-utils and node writeFile functions.

## Checklist

- [ ] Tests added


packages/server-dev/src/manager/processes/nextBuild.mjs
```diff
@@ -16,14 +16,14 @@
...
   return async () => {
     console.log('Building app...');
     await spawnProcess({
       logger: console,
-      args: ['run', 'build:next'],
-      command: packageManager,
-      silent: !verbose,
+      command: 'node',
+      args: [context.bin.next, 'build'],
+      silent: !context.verbose,
```
Gervwyk, line 26: Let's rename `silent` to `verbose` or other way around. Renaming params here and switching to inverse, yet in spawnProcess you do  `if (!silent)`.

packages/server-dev/src/manager/processes/nextBuild.mjs
```diff
@@ -16,14 +16,14 @@
 
 import { spawnProcess } from '@lowdefy/node-utils';
 
-function nextBuild({ packageManager, verbose }) {
+function nextBuild(context) {
   return async () => {
     console.log('Building app...');
```
Gervwyk, line 21: Add `TODO` to pass down logger to these methods to use and pass through to spawnProcess.

packages/server-dev/src/manager/processes/shutdownServer.mjs
```diff
@@ -15,10 +15,19 @@
...
-    if (context.serverProcess) {
-      console.log('Shutting down server...');
-      context.serverProcess.kill();
+  return () => {
+    if (context.nextServer) {
+      // console.log(
+      //   `Existing server ${context.nextServer.pid}, killed: ${context.nextServer.killed}`
+      // );
+      if (!context.nextServer.killed) {
+        console.log('Shutting down server...');
```
Gervwyk, line 24: Maybe later change to something like `logger.verbose()`, `logger.log()` for these additional logs.

packages/server-dev/src/manager/processes/startWatchers.mjs
```diff
@@ -14,50 +14,17 @@
   limitations under the License.
 */
 
-import configWatcher from '../watchers/configWatcher.mjs';
 import envWatcher from '../watchers/envWatcher.mjs';
-
```
Gervwyk, line 19: This breakdown of the process was nice to include / understand. Maybe we can keep / maintain it somewhere?

packages/server-dev/src/manager/watchers/nextBuildWatcher.mjs
```diff
@@ -0,0 +1,87 @@
...
+
+  const callback = async (filePaths) => {
+    let install = false;
+    let build = false;
+    await Promise.all(
+      filePaths.flat().map(async (filePath) => {
+        const hash = await sha1(filePath);
+        if (hashes[filePath] === hash) return;
+        build = true;
+        if (filePath.endsWith('package.json')) install = true;
```
Gervwyk, line 62: We've always been writing `if () {}` instead of short hand for readability.

Gervwyk (approved):

PR #1094, opened 2022-02-08
Add Lowdefy Link component for blocks.

Gervwyk:
Closes ~#ISSUE_NUMBER~

## Checklist

- [x] Tests added


packages/build/src/build/validateConfig.js
```diff
@@ -39,6 +39,11 @@ async function validateConfig({ components }) {
   if (type.isNone(components.config.theme)) {
     components.config.theme = {};
   }
+  if (type.isString(components.config.basePath)) {
```
SamTolmay, line 42: This is good 👍

packages/engine/src/createLink.js
```diff
@@ -16,32 +16,42 @@
...
+      // Cannot set input or urlQuery on back
+      return backLink(props);
     }
-    if (!type.isNone(url)) {
-      return newOriginLink(`${url}${lowdefyUrlQuery}`, newTab);
+    const lowdefyUrlQuery = type.isNone(props.urlQuery)
+      ? ''
+      : `?${urlQueryFn.stringify(props.urlQuery)}`;
+    if (props.home === true) {
+      lowdefy.inputs[`page:${lowdefy.home.pageId}`] = props.input;
```
SamTolmay, line 37: So this will set input to undefined or null if it isn't defined here, but it was set before. I think that is the expected behaviour, but this is a breaking change from:
```js
      if (!type.isNone(input)) {
         lowdefy.inputs[pageId] = input;
       }
```
Maybe we should also set a empty object default here, or ensure the null is correctly handled in engine because I don't think engine checks this except when the context is created

packages/plugins/blocks/blocks-antd/src/blocks/Menu/Menu.js
```diff
@@ -288,16 +267,13 @@ const MenuComp = ({
...
-                                  basePath={basePath}
-                                  Link={Link}
-                                  linkStyle={methods.makeCssClass(subLinkGroup.style, true)}
-                                  id={subLinkGroup.id}
-                                  makeCssClass={methods.makeCssClass}
-                                  pageId={subLinkGroup.pageId}
-                                  properties={subLinkGroup.properties}
-                                  url={subLinkGroup.url}
-                                />
+                                <Link
```
SamTolmay, line 270: So will menus now support `urlQuery` and `input`?

packages/plugins/blocks/blocks-basic/src/blocks/Anchor/Anchor.js
```diff
@@ -15,28 +15,9 @@
...
-    <span id={blockId} className={className}>
-      {children}
-    </span>
-  ) : (
-    <Link
-      id={blockId}
-      className={className}
-      href={href}
-      onClick={onClick}
-      rel={rel || 'noopener noreferrer'}
```
SamTolmay, line 33: I assume these are set properly by `next/link`?

packages/server/src/components/createLinkComponent.js
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import NextLink from 'next/link';
+import { createLink } from '@lowdefy/engine';
+import { type } from '@lowdefy/helpers';
+
+const createLinkComponent = (lowdefy) => {
+  const backLink = ({ children, className, id }) => (
+    <a id={id} onClick={() => lowdefy._internal.router.back()} className={className}>
```
SamTolmay, line 8: Does this need anything for a11y?

packages/engine/src/createLink.js
Gervwyk: I wouldn't set it here, but rather default during read, which is think is handled by `get`. When looking at `lowdefy.inputs` it's nice to see only inputs that were set explicitly. Else we might default to an empty object and make it hard to inspect why the value was not passed down. In short, undefined is better for debugging in my opinion. Thoughts?

packages/plugins/blocks/blocks-antd/src/blocks/Menu/Menu.js
Gervwyk: Amazing right 😱 should I update the docs?

packages/plugins/blocks/blocks-basic/src/blocks/Anchor/Anchor.js
Gervwyk: createLink handles them where applicable, but should probably be configurable.

packages/server/src/components/createLinkComponent.js
Gervwyk: I'll add some aria-labels.

SamTolmay (approved):

PR #1113, opened 2022-02-15
Server cli scripts

Gervwyk:
Closes ~#ISSUE_NUMBER~

Refine scripts to specify server and dev server directories to cli to also run app and docs through cli.

## Checklist

- [x] Tests added


packages/build/src/build/buildRefs/runTransformer.js
```diff
@@ -22,7 +22,12 @@ async function runTransformer({ context, parsedFile, refDef }) {
       context,
       filePath: refDef.transformer,
     });
-    return transformerFn(parsedFile, refDef.vars);
+    try {
+      return transformerFn(parsedFile, refDef.vars);
+    } catch (error) {
+      context.logger.error(`Error running transformer ${refDef.transformer} for ${refDef.path}.`);
```
SamTolmay, line 28: We do :
```js
  } catch (error) {
    throw new Error(
      `Error calling resolver "${refDef.resolver}" from "${referencedFrom}": ${error.message}`
    );
  }
```
in `runRefResolver`.

Maybe we should be consistent between the two. I prefer the message here with the quotes, but I am not sure if doing context.logger.error or just throwing results in a better looking error for the user

SamTolmay (approved):

PR #1137, opened 2022-03-01
fix(actions-core): Fix Throw action error message if params are not a…

SamTolmay:
…n object.

Closes ~#ISSUE_NUMBER~

The throw action now has a better error message if the params are not an object.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/plugins/actions/actions-core/src/actions/Throw.js
+++ b/packages/plugins/actions/actions-core/src/actions/Throw.js
@@ -27,6 +27,9 @@ class ThrowActionError extends Error {
 }
 
 function Throw({ methods: { getBlockId, getPageId }, params }) {
+  if (!type.isObject(params)) {
+    throw new Error(`Invalid Throw, check action params. Received "${JSON.stringify(params)}".`);
+  }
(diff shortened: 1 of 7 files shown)


packages/plugins/actions/actions-core/src/actions/Throw.js
```diff
@@ -27,6 +27,9 @@ class ThrowActionError extends Error {
 }
 
 function Throw({ methods: { getBlockId, getPageId }, params }) {
+  if (!type.isObject(params)) {
+    throw new Error(`Invalid Throw, check action params. Received "${JSON.stringify(params)}".`);
```
Gervwyk, line 31: If the test throws for params, not an object, can we change the error message to say that params must be an object?

SamTolmay: Yes we can. I originally refactored so that we only have one throw, but then the test where the `throw` property is a string does not throw like before. I think we should definitly ensure that `throw` is a boolean or none.

I will update this so that those are checked better, with better error messages.

Gervwyk (approved):

PR #1187, opened 2022-05-01
Client reload fixes and display message

Gervwyk:
- Add a reset context flag to client so server-dev watcher can tell client to recreate the context.
- Bug fixes to ensure that onInit fires before onMount starts, and render progress bar until final onMount has completed.
- Also add display message implementation.

## Todo: 
- Reload needs to pass a flag to client when server is restarting so the client does not crash when new types are added to the server. This flag can also be used to render a "server restarting" screen, for now we delay render as a hack.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/client/src/Client.js
+++ b/packages/client/src/Client.js
@@ -18,6 +18,7 @@ import React from 'react';
 
 import Block from './block/Block.js';
 import Context from './Context.js';
+import DisplayMessage from './DisplayMessage.js';
 import Head from './Head.js';
 import ProgressBarController from './ProgressBarController.js';
 
@@ -36,10 +37,20 @@ const Client = ({
   return (
     <>
       <ProgressBarController
-        id="page-loader"
-        key={`${config.pageConfig.id}-loader`}
+        id="lowdefy-progress-bar"
+        key={`${config.pageConfig.id}-progress-bar`}
         lowdefy={lowdefy}
         resetContext={resetContext}
+      />
+      <DisplayMessage
+        id="lowdefy-displayy-message"
+        key={`${config.pageConfig.id}-display-message`}
+        Component={lowdefy._internal.blockComponents.Message}
+        methods={{
+          registerMethod: (_, method) => {
+            lowdefy._internal.displayMessage = method;
+          },
+        }}
       />
       <Context
         key={config.pageConfig.id}
--- a/packages/client/src/Client.js
+++ b/packages/client/src/Client.js
@@ -43,7 +43,7 @@ const Client = ({
         resetContext={resetContext}
       />
       <DisplayMessage
-        id="lowdefy-displayy-message"
+        id="lowdefy-display-message"
         key={`${config.pageConfig.id}-display-message`}
         Component={lowdefy._internal.blockComponents.Message}
         methods={{
--- a/packages/client/src/Client.js
+++ b/packages/client/src/Client.js
@@ -23,16 +23,30 @@ import ProgressBarController from './ProgressBarController.js';
 
 import initLowdefyContext from './initLowdefyContext.js';
 
-const Client = ({ Components, config, router, stage, types, window }) => {
-  const lowdefy = initLowdefyContext({ Components, config, router, types, stage, window });
+const Client = ({
+  Components,
+  config,
+  resetContext = { reset: false, setReset: () => undefined },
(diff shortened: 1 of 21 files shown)


packages/client/src/Client.js
```diff
@@ -18,47 +18,66 @@ import React from 'react';
...
 
-const Client = ({ Components, config, router, stage, types, window }) => {
-  const lowdefy = initLowdefyContext({ Components, config, router, types, stage, window });
-
+const Client = ({
+  Components,
+  config,
+  resetContext = { reset: false, setReset: () => undefined },
+  router,
+  stage,
```
SamTolmay, line 32: Do we still use `stage` with resetContext?

packages/client/src/Client.js
```diff
@@ -18,47 +18,66 @@ import React from 'react';
...
-    />
+    <>
+      <ProgressBarController
+        id="lowdefy-progress-bar"
+        key={`${config.pageConfig.id}-progress-bar`}
+        lowdefy={lowdefy}
+        resetContext={resetContext}
+      />
+      <DisplayMessage
+        id="lowdefy-displayy-message"
```
SamTolmay, line 46: lowdefy-display-message

packages/engine/src/Blocks.js
```diff
@@ -91,8 +91,13 @@ class Blocks {
       block.styleEval = {};
       block.validationEval = {};
       block.visibleEval = {};
-
-      block.meta = this.context._internal.lowdefy._internal.blockComponents[block.type].meta;
+      try {
```
SamTolmay, line 94: Should engine be the thing throwing here? I guess we create the context before we start rendering, so this is the first place it throws. But build is also not supposed to build if a block type doesn't exist.

Is this the right error message? Why were you running into this error without build throwing?

One other thing that could happen here is a bad block that does not have a meta definition, or a plugin that declares a block that does not exist.

packages/server-dev/lib/utils/waitForRestartedServer.js
```diff
@@ -26,7 +26,7 @@ function waitForRestartedServer(basePath) {
     } catch (error) {
       waitForRestartedServer(basePath);
     }
-  }, 1500);
+  }, 500); // TODO: this ping should be shorter than rerender delay until we can pass a rebuild flag to reload.
```
SamTolmay, line 29: Not sure what you mean here?

packages/server-dev/lib/RestartingPage.js
```diff
@@ -0,0 +1,46 @@
...
+    >
+      <div
+        style={{
+          verticalAlign: 'middle',
+          display: 'inline-block',
+        }}
+      >
+        <h3>Rebuilding Lowdefy App</h3>
+        <p>
+          The server is reconfiguring Lowdefy types and restarting. The page will reload when ready.
```
SamTolmay, line 39: Nice! This is a good idea.

This could also happen when app config changes, or .env changes, so maybe `reconfiguring Lowdefy types` is not the right message.

Maybe we could do something like `The server is restarting because your configuration changed. The page will reload when ready.`

As an idea, maybe one day we can display some Lowdefy tips and tricks here, like games do with loading screens.

packages/client/src/Client.js
Gervwyk: Yes, to add lowdefy to window.

packages/engine/src/Blocks.js
Gervwyk: True, those checks should happen else where as well, however, this error is good to indicate that there is an issue with the block, instead of simply throwing can not read meta of undefined. Perhaps the message should only read: `Invalid block definition for type ${type} at block ${blockId}`?

packages/server-dev/lib/RestartingPage.js
Gervwyk: Cool. Changed the message. Yeah, like the tips idea. Would like to fetch that from the telemetry server. Please add to ideas channel.

packages/engine/src/Blocks.js
SamTolmay: Happy with this message, can we add it in the next PR? Or this oneif you make other changes.

SamTolmay (approved):

PR #1249, opened 2022-06-24
fix(client): Skeleton to get parent blocks properties and styles.

Gervwyk:
If no properties of styles are defined on the skeleton definition, it will default the properties and style values of the parent blocks.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/client/src/block/LoadingContainer.js
+++ b/packages/client/src/block/LoadingContainer.js
@@ -38,7 +38,7 @@ const LoadingContainer = ({
         area={layoutParamsToArea({
           area: skeleton.areas[areaKey] || {},
           areaKey,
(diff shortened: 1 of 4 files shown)


packages/client/src/block/LoadingContainer.js
```diff
@@ -29,7 +38,7 @@ const LoadingContainer = ({ blockId, Component, context, layout, lowdefy, skelet
         area={layoutParamsToArea({
           area: skeleton.areas[areaKey] || {},
           areaKey,
-          layout,
+          layout: blockLayout ?? {},
```
SamTolmay, line 41: Shouldn't this have `skeleton.layout` as well?

SamTolmay (approved):

PR #1282, opened 2022-07-14
JS Operators

SamTolmay:
Closes ~#ISSUE_NUMBER~

This PR adds the `_intl` operator, and adds the `_object.entries` and `_object.fromEntries` methods. It also improves the handling of null values in the `_array` and `_object` operators.

The `_intl` operator replaces functionality that was previously part of the `_format` operator.

## Checklist

- [x] Tests added


packages/plugins/operators/operators-js/src/operators/shared/array.js
```diff
@@ -76,7 +76,7 @@ const meta = {
     prep,
     validTypes: ['array', 'object'],
   },
-  reverse: { prep, validTypes: ['array'], singleArg: true },
+  reverse: { prep, validTypes: ['array', 'null'], singleArg: true },
```
Gervwyk, line 79: Why are we only changing reverse and not filter, map, etc.. And if null is the input, should and array be the output or simply `null`?

Gervwyk (approved):

PR #1303, opened 2022-07-22
Implementation for Lowdefy config traces for client errors [Part 1]

Gervwyk:
The first step for better dev-ex is to throw more informative errors. This PR adds config traces to all app objects in order to links any app object to its config origin or reference definition. 

In doing this we are implementing a breaking change. Previously operators would fail silently and only log to browser console. For logging, we'll be logging all errors, and having operators log errors but not break will result a lot of errors which are ignored, which results in a bad quality log stream. Making operator errors breaking will result in making the operators more robust as well as users writing more robust config. 

## Checklist

- [x] Tests added


Gervwyk:
Closing this PR for now. We will continue implementation in v4 and reimplement. These changes are currently a high-priority item on the roadmap.

PR #1305, opened 2022-07-22
Update Connections and Requests concept page for v4.

StephanieJKS:
Closes ~#ISSUE_NUMBER~

Updates the Connections and Request concept page to be relevant for v4.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/docs/concepts/connections-and-requests.yaml
+++ b/packages/docs/concepts/connections-and-requests.yaml
@@ -24,13 +24,7 @@ _ref:
         type: Markdown
         properties:
           content: |
-            ### TLDR
-              - `connections` define links to other services, like connecting to a database. They are defined at the root of the lowdefy configuration.
-              - `requests` use connections to make a call to the connected external services.
-              - Use the [`_secret`](/_secret) operator to reference API keys or other secrets as required - do not code secrets into your config or commit secrets to your config source control.
-
-            -----------
-            In a Lowdefy app you can integrate with other services like API's or databases using `connections` and `requests`. Connections configure the connection settings to the service, and often contain parameters like connection strings, urls and secrets like passwords or API keys. Requests are used to interact with the connection, like inserting a data record, executing a query or calling a API end-point.
+            In a Lowdefy app you can integrate with other services like API's or databases using `connections` and `requests`. Connections configure the connection settings to the service, and often contain parameters like connection strings, urls and secrets like passwords or API keys. Requests are used to interact with the connection, such as inserting a data record, executing a query or calling a API end-point.
 
       - id: alert2
         type: Alert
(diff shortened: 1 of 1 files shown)


packages/docs/concepts/connections-and-requests.yaml
```diff
@@ -43,12 +37,16 @@ _ref:
         type: Markdown
         properties:
           content: |
+            To implement requests, the following steps are required:
+            - define a connection.
+            - define a request.
+            - call the request using a [request action](/Request).
+            - use returned data by making use of the [request operator](/_request).
```
SamTolmay, line 44: I think thes should be `[Request](/Request) action` and `[_request](/_request)operator`.

packages/docs/concepts/connections-and-requests.yaml
```diff
@@ -75,16 +73,28 @@ _ref:
 
             # Requests
 
-            Requests can be defined on any block, and the results of the request are available to any block in the same context. Requests must have an `id`, `type`, `connectionId` field specifying the connection to use, and `properties` defining the request settings. Requests can be called using the [`Request`](/Request) action. Operators in request properties are evaluated every time a request is called.
+            Requests can be defined on any block, and the results of the request are available to any block in the same context. Requests must have an `id`, `type`, `connectionId` field specifying the connection to use, and `properties` defining the request settings. Requests can be called using the [`Request`](/Request) action.
```
SamTolmay, line 76: `and the results of the request are available to any block in the same context` should be removed  and replaced with some like `the results of the request can be access using the _request operator anywhere on the same page`.

packages/docs/concepts/connections-and-requests.yaml
```diff
@@ -24,30 +24,28 @@ _ref:
...
-          message: Sensitive information like passwords or API keys are often required to use external services. The _secret operator should be used to reference these secrets, they should never be coded directly in your app, or committed to source control.
+          message: Sensitive information like passwords or API keys are often required to use external services. The <a href = "./_secret">_secret</a> operator should be used to reference these secrets, they should never be coded directly in your app, or committed to source control.
 
       - id: md2
         type: Markdown
         properties:
           content: |
-            # Connections
+            To implement requests, the following steps are required:
+            - define a connection.
```
SamTolmay, line 41: Should these start with capitals?

packages/docs/concepts/connections-and-requests.yaml
```diff
@@ -73,18 +71,30 @@ _ref:
...
 
-            # Requests
+            ## Requests
+
+            Requests can be defined on any block, and the results of the request can be access using the `_request` operator anywhere on the same page. Requests must have an `id`, `type`, `connectionId` field specifying the connection to use, and `properties` defining the request settings. Requests can be called using the [`Request`](/Request) action.
 
-            Requests can be defined on any block, and the results of the request are available to any block in the same context. Requests must have an `id`, `type`, `connectionId` field specifying the connection to use, and `properties` defining the request settings. Requests can be called using the [`Request`](/Request) action. Operators in request properties are evaluated every time a request is called.
+            The client operators to be used are defined in the `payload` field and are accessed by making use of the `_payload` operator in the `properties` field. These operators are evaluated on the client. Operators defined in `properties`, such as `_secret` and `_user`, are evaluated on the server. Operators in the request are evaluated every time a request is called.
```
SamTolmay, line 78: Maybe we could do something like:
```
Data that should be passed to the request from the client can be defined in the `payload` field. Operators can be used to create the payload. An example of this would be to use the `_state` operator to include the value of an input block in the payload. The data in the payload can accessed in the request by making use of the `_payload` operator in the `properties` field. The operators defined in the properties field are evaluated on the server. Operators in the request are evaluated every time a request is called.
```

SamTolmay (approved):

packages/docs/concepts/connections-and-requests.yaml
StephanieJKS: I think they should. 

Change made in 97d1d764505e2c23bfcbf67116271b2ab8a853f0

packages/docs/concepts/connections-and-requests.yaml
StephanieJKS: This makes sense.

Change made in 97d1d764505e2c23bfcbf67116271b2ab8a853f0

PR #1309, opened 2022-07-25
Update Events and Actions concept page for v4.

StephanieJKS:
Closes ~#ISSUE_NUMBER~

General improvements made to the Events and Actions concept page.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/docs/concepts/events-and-actions.yaml
+++ b/packages/docs/concepts/events-and-actions.yaml
@@ -24,41 +24,25 @@ _ref:
         type: MarkdownWithCode
         properties:
           content: |
-            ### TLDR
-            - Events are triggered when something happens on a page, like clicking a button or loading a page.
-            - A list of actions are executed sequentially by a triggered event.
-            - If an action errors, the actions that follow are skipped.
-            - Actions that are `async: true` will not executed sequentially nor stop the event if they error.
-            - Action errors can be handled by providing a list of `try` and `catch` actions to the event.
-            - Operators used in action `params` are evaluated right before the action is executed.
-            - The [`_actions`](/_actions)) operator is available for sequential actions to use the values returned from preceding actions in the chain.
-            - Actions have a `skip` field that can be used to skip action execution.
-            - The `onInit` event is triggered the first time a context is mounted and keeps the page in loading until all actions have finished.
-            - The `onInitAsync` event is triggered the first time a context is mounted and does not keep the page in loading.
-            - The `onMount` events is triggered the every time a block is mounted and keeps the block in loading until all actions have finished.
-            - The `onMountAsync` event is triggered the every time a block is mounted, after `onMount` has completed, and does not keep the block in loading.
-
-            -----------
-
             Blocks can define _events_ which the block can trigger when something happens on the page, like a button being clicked, an input's value being modified or a page being loaded. Some examples are `onClick` on a [`Button`](/Button) or `onMount` on a [`PageHeaderMenu`](/PageHeaderMenu) block.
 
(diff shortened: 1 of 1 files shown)


packages/docs/concepts/events-and-actions.yaml
```diff
@@ -24,41 +24,25 @@ _ref:
...
 
-            Actions can have a `params` field for specifying input parameters when executing the action. Operators used in action `params` will be evaluated right before the action is executed. Some events might have data relating to that event, like what the new value of an input is, or the row that was clicked in a table. The `event` object can be used in the action using the [`_event`](/_event) operator. Some actions also return values which can be passed to preceding actions in the same action chain using the [`_actions`](/_actions) operator.
+            Actions can have a `params` field for specifying input parameters when executing the action. Operators used in action `params` will be evaluated right before the action is executed. Some events might have data relating to that event, such as what the new value of an input is, or the row that was clicked in a table. The `event` object can be used in the action using the [`_event`](/_event) operator. Some actions also return values which can be passed to preceding actions in the same action chain using the [`_actions`](/_actions) operator.
```
SamTolmay, line 37: Should be succeding instead of preceding actions.

packages/docs/concepts/events-and-actions.yaml
```diff
@@ -24,41 +24,25 @@ _ref:
...
 
-            _Actions_ are tasks that can be executed, like calling a request, linking to a new page or changing a value in state. An array of actions can be defined for a event on a block. If that event gets triggered, those actions will execute sequentially. If any actions error while executing, the actions that follow it won't be executed, however, `catch` actions chain can be defined on a event to trigger when an error in a chain of actions occurs.
+            _Actions_ are tasks that can be executed, like calling a request, linking to a new page or changing a value in state. An array of actions can be defined for an event on a block. If that event gets triggered, those actions will execute sequentially. If any actions error while executing, the actions that follow it won't be executed, however, `catch` actions chain can be defined on an event to trigger when an error in a chain of actions occurs.
```
SamTolmay, line 31: I think we can update this to something like this:
```
_Actions_ are tasks that can be executed, like calling a request, linking to a new page or changing a value in state. An array of actions can be defined for an event on a block. If that event gets triggered, those actions will execute sequentially. 

The chain of actions will stop executing if the action raises a break condition (what do we call this?) (eg. when a form validation does not pass) or if the action throws an unexpected error.

If a break condition was raised, the user will be shown an error message and no further actions will execute. If an unexpected error is throw the app will show an error screen and the error will be logged. To handle errors and break conditions, a `catch` action chain can be defined on the event to handle the error, as described below.
```
@Gervwyk please check and confirm the described behaviour

packages/docs/concepts/events-and-actions.yaml
```diff
@@ -194,7 +178,7 @@ _ref:
 
             The `onMountAsync` event is triggered every time a block is mounted, but does not render the block in loading.
 
-            # Action types
+            ## Action types
 
             The following actions can be used:
```
SamTolmay, line 184: `Message` was renamed to `DisplayMessage`

 The following need to be removed:
JsAction
Notification

The following actions need to be added:
Login
Logout
Fetch
ResetValidation
Break
Wait

Please update these links, and then update the individual action docs so the links match up. I am happy if you leave TODOs for the actions that don't have docs.

For Break @Gervwyk can update the Break action docs as part of his changes.

SamTolmay (approved):

packages/docs/concepts/events-and-actions.yaml
StephanieJKS: Change made in eb614758df9c5036d24ce224142dda3c60172021

packages/docs/concepts/events-and-actions.yaml
StephanieJKS: Changed this to be similar to what you have suggested there, for now.

Change made in https://github.com/lowdefy/lowdefy/commit/eb614758df9c5036d24ce224142dda3c60172021

PR #1314, opened 2022-07-29
Update Lowdefy App Schema concept page.

StephanieJKS:
Closes ~#ISSUE_NUMBER~

Makes changes to the Lowdefy App Schema concept page. These changes are made to generally improve the page, and make it more relevant to v4.

## Checklist

- [ ] Tests added


packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -53,9 +53,10 @@ _ref:
             - `pages: object[]`: An array of page objects.
 
             Pages are made up of blocks. Blocks have the following basic schema:
-            - `id: string`: __Required__ - A unique identifier for a block.
-            - `type: string`: __Required__ - This is the block type identifier and defines what block to used.
-            - `properties: object`: All the settings passed to a block component.
+            - `id: string`: __Required__ - An identifier for a block.
```
SamTolmay, line 56: Block id is no longer required

packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -80,21 +81,19 @@ _ref:
             The config object has the following properties:
 
             - `basePath: string`: Set the base path to serve the Lowdefy application from. This will route all pages under `https://example.com/<base-path>/<page-id>` instead of the default `https://example.com/<page-id>`. The basePath value must start with "/".
-            - `homePageId: string`: The pageId of the page that should be loaded when a user loads the app without a pageId in the url route. This is the page that is loaded when you navigate to `yourdomain.com`.
+            - `homePageId: string`: The id of the page that will load when a visitor navigates to the home route, for example, `yourdomain.com`.
```
SamTolmay, line 84: I think the "for example, `yourdomain.com`." part might be a bit misleading, which is why the original explanation is so verbose. This is specifically if the user navigates to `example.com` or `example.com/`, not `example.com/page-id`.

packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -80,21 +81,19 @@ _ref:
...
 
             ## Connections
 
-            In a Lowdefy app you can integrate with other services like API's or databases using `connections` and `requests`. Connections configure the connection settings to the service, and often contain parameters like connection strings, urls and secrets such as passwords or API keys. Requests are used to interact with the connection, for example inserting a data record, executing a query or calling a API end-point.
+            In a Lowdefy app you can integrate with other services like API's or databases using `connections` and `requests`. Use connections to configure the settings to a service, for example, settings such as urls, authentication config or [`secrets`](/secrets) like passwords or API keys. Requests use connections to send and receive data from external services, for example, inserting a data record or executing a query on a database, or simply calling an API end-point.
```
SamTolmay, line 96: This is nice.

Maybe the 
"settings such as urls, authentication config or [`secrets`](/secrets) like passwords or API keys." can be change to "settings such as urls or [`secrets`](/secrets) like passwords or API keys." since the authentication config is a little redundent and i thing the other two are clear enough.

SamTolmay, line 96: We can maybe do something a little better than 
"send and receive data from external services,"
To me this is more specific actions with the external service API, so i like the word "interactions" there. The request id the part of the connection that does the work. With custom plugins this can become quite complicated as well, you could write a whole piece of code that executes business logic there.

I guess it all boils down to sending and receiving data, but I don't think that captures the full idea

SamTolmay: Some requests like sending an email or processing a payment with Stripe might not feel like "sending data" to general users.

packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -104,21 +103,21 @@ _ref:
 
             ## Menus
 
-            Menu objects are objects with links to pages. They are filtered to only show pages that the user is authorized to see and are used by blocks such as `PageSiderMenu` to render the links in the menu. If no menus are provided, a default menu is created, with links to all of the defined pages, and with their pageIds as menu item titles.
+            Menu objects describe links to pages, within the app or external. Menu lists are filtered to only show pages that the user is authorized to see as a result of public, private or [role based access controlled (RBAC)](/roles) pages. Blocks such as `PageSiderMenu` render menu links and by default if no menu defined, a default menu is created, containing links to all pages defined in the app.
```
SamTolmay, line 106: This is nice.

I would split the default menu sentance to a new sentance:
`render menu links. If no menu is defined, a default menu is created, containing links to all pages defined in the app.`

packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -157,11 +154,9 @@ _ref:
                 blocks:
                   - id: content_card
                     type: Card
+                    properties:
```
SamTolmay, line 157: In the example description we say we are going to add a title block. We should probably just adjust that sentance

packages/docs/concepts/lowdefy-schema.yaml
```diff
@@ -273,15 +264,17 @@ _ref:
 
             ## References and templates
 
-            References are used to split the configuration of an app into logically distinct files, and to reuse configuration in the app. References can be used almost anywhere in the configuration, as long as the configuration remains valid YAML.
+            References are used to split the configuration of an app into logically distinct files, and to reuse configuration in the app. Configuration is primarily split up for two reasons, to improve readability or for reusability, for example, when used as templates. References can be used anywhere in the configuration, as long as the configuration resolves to comply with the Lowdefy app schema.
```
SamTolmay, line 267: When using references there are the following cases:
- Just splitting out the config for readabilty and code navigation
- Splitting out a piece of config because you want to use the exact same config in multiple places (code reuse level 1)
- Using a ref with variables so you can have a shared piece of config with some dynamic variables (code reuse level 2)
- Using an njk file which allows for much more complicated things - this is an advanced feature and easily overused (code reuse level 3).
- Using transformers - even more compplicated than njk files - should be used if justified like building config from a JSONSchema file (code reuse level 4).

I think the current explanation does not elaborate on the level 1 and 2 enough, and it feels like we jump directly to nunjucks templates. I also don't know if `templates` is the right word for level 1 and 2 code reuse, becuase of the confusion with nunjucks.

SamTolmay (commented):
Very nice, these changes really help with readability.

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: Should we go into more detail on that here or on the References and Templates concept page?

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: Change made in 386463869c78c494c05636e1e435419c96917b02 
Tried to get a bit more of a balance between the two versions.

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: Change made in 386463869c78c494c05636e1e435419c96917b02

StephanieJKS: Not sure if changing "send and receive data from external services" to "interact with external services" is better here. That's the current change that I have made in the latest commit. Let me know if you have better ideas.

Change made in 386463869c78c494c05636e1e435419c96917b02

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: This definitely reads better.
Change made in 386463869c78c494c05636e1e435419c96917b02

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: I have changed up the example description. Let me know what you think and if it can be improved or should be worded differently.
Change made in 386463869c78c494c05636e1e435419c96917b02

packages/docs/concepts/lowdefy-schema.yaml
Gervwyk: I like the breakdown @SamTolmay. I'd consider listing this description on both pages? It's important to know it exists, and can be hard to come to this realization when you are building apps. We're also not making it easy since the same tool / named thing can be used to solve all these probelms.

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: Change made in https://github.com/lowdefy/lowdefy/commit/386463869c78c494c05636e1e435419c96917b02

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: Made some changes to this section in 5837c29b0e95dd7432b8f4c6d314179c9aaf1428

Gervwyk (approved):

packages/docs/concepts/lowdefy-schema.yaml
StephanieJKS: > Made some changes to this section in [5837c29](https://github.com/lowdefy/lowdefy/commit/5837c29b0e95dd7432b8f4c6d314179c9aaf1428)

Made the same changes in 1ee298ee370bac30e1189237d5cf9ccb69a5fac3

PR #1350, opened 2022-08-24
Fix label validation style render bug.

StephanieJKS:
Closes ~#ISSUE_NUMBER~

This PR fixes the bug in which the validate styles were not being applied correctly. These included border color not changing; color of feedback in extra not changing; and the icon not being displayed in the correct position, as well as not changing color. 
The fix for the component border color involved setting the status property of the relative components. 
The fix for the feedback in the extra involved changing how the classes were applied.
The fix for the icon involved changing the classes applied to the icon.

## Checklist

- [ ] Tests added


packages/plugins/blocks/blocks-antd/src/blocks/DateRangeSelector/DateRangeSelector.js
```diff
@@ -51,7 +51,7 @@ const DateRangeSelector = ({
       validation={validation}
       required={required}
       content={{
-        content: () => (
+        content: (icon) => (
```
Gervwyk, line 54: Can we add an TODO here to say: Content area should pass down named args.

Gervwyk (approved):

packages/plugins/blocks/blocks-antd/src/blocks/DateRangeSelector/DateRangeSelector.js
Gervwyk: Actually, let's change these to `({ icon }) => ` and call with `({ icon })`, please.

StephanieJKS: Change made in 832f68c671e23c30f960792c8dd91164c6742f6a

Gervwyk (dismissed):

Gervwyk (approved):

PR #1458, opened 2023-01-06
Lowdefy Website

StephanieJKS:
This PR adds the Lowdefy website as a new package. This website package recreates the existing Lowdefy website using lowdefy config.

## Checklist

- [ ] Tests added


packages/website/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "./types": "./dist/types.js"
+  },
+  "files": [
+    "dist/*"
+  ],
+  "scripts": {
+    "build": "swc src --out-dir dist --config-file ../../.swcrc --delete-dir-on-start",
+    "clean": "rm -rf .lowdefy",
+    "prepublishOnly": "pnpm build",
+    "test": "node --experimental-vm-modules node_modules/jest/bin/jest.js"
```
SamTolmay, line 46: Please remove these scripts as well as `dependencies` and `devDependencies` and `exports`, because they are not used. Unless we are planning to use custom plugins for the website, in which case they should stay

Removing the test script will make the github workflows tests pass, since Jest fails if it doesn't find any tests

SamTolmay: We actually maybe don't need a `package.json` at all if we are not using plugins, this might be better because then then we won't publish an empty package to npm

StephanieJKS: We might need a custom plugin or two for a few of the components that still need to be added. 

Please see 5cfff352b93631dd86ec75f7dab0aed09370df31

SamTolmay (dismissed):

SamTolmay (approved):

PR #1478, opened 2023-01-20
Authentication Docs

SamTolmay:
Closes #ISSUE_NUMBER

## Checklist

- [ ] Tests added


packages/docs/users/roles.yaml
```diff
@@ -18,53 +18,52 @@ _ref:
...
         properties:
-          content: |
-            Roles can be used to limit user access to certain pages. Only users with the role will be able to see the page, and the page will be filtered from menus if the user does not have the role.
+          content:
+            _nunjucks:
+              on:
+                version:
+                  _ref: version.yaml
+              template: |
+                Roles can be used to limit user access to certain pages. Only users with the role will be able to see the page, and the page will be filtered from menus if the user does not have the role.
```
StephanieJKS, line 32: Sentence can be a bit confusing. What 'role' are we referring to? 

Maybe something like 'Only users with the role linked to the page will be able to see that page...'

packages/docs/users/login-and-logout.yaml
```diff
@@ -104,61 +111,45 @@ _ref:
...
+                      urlQuery:
+                        screen_hint: signup
             ```
 
             ## Logout
 
-            When the `Logout` action is called, the user data and authorization cookie are cleared by the app. The `Logout` action does not take any parameters. The user is then redirected to the URL configured in `logoutRedirectUri`, or the app homepage if this is not configured.
+            When the `Logout` action is called, the user data and authorization cookies are cleared by the app.
+
+            The `callbackUrl` parameters of the Logout action specify where the user is redirected after logout is complete. If `callbackUrl` parameters not set, the user will return to the page from which the action was called.
```
StephanieJKS, line 123: If callbackUrl parameters `are` not set...

StephanieJKS (commented):
Overall looks good. Easy to read and explains concepts quite well.

Gervwyk (approved):

PR #1480, opened 2023-01-24
Throw error if CLI version 3 and app version > 3

StephanieJKS:
This PR adds an error that is thrown when a user tries to run an app with a version greater than 3 on the version 3 CLI. It adds a dependency on semver, which is used to validate the version and to get the major version.

## Checklist

- [ ] Tests added


packages/cli/src/utils/checkForUpdatedVersions.js
```diff
@@ -15,8 +15,24 @@
...
+---------------------------------------------------
+  Version ${lowdefyVersion} is not a valid version.
+---------------------------------------------------`);
+  }
+  if (semver.major(lowdefyVersion) > 3) {
+    throw new Error(`
+---------------------------------------------------
+  You are attempting to run a version ${lowdefyVersion} app with the version 3 CLI.
+  Please update your Lowdefy CLI version.
+  To do this, run 'pnpx lowdefy@${semver.major(lowdefyVersion)}'.
```
SamTolmay, line 32: This should be `npx` for version 3.

SamTolmay (approved):

PR #1531, opened 2023-03-14
Add GeolocationCurrentPosition action to the docs.

StephanieJKS:
This PR adds the GeolocationCurrentPosition action to the docs. It also updates the action template to add a response section if `response` is provided to the template.

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/docs/actions/GeolocationCurrentPosition.yaml
+++ b/packages/docs/actions/GeolocationCurrentPosition.yaml
@@ -1,0 +1,61 @@
+# Copyright 2020-2023 Lowdefy, Inc
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


SamTolmay (dismissed):

Gervwyk (changes requested):
I think we need to implement better error handling for when geolocation is not available or denied by client. Currenlty we wrap it in a try catch. and return null is it fails. Perhaps not throw to break the action chain but instead return an adequate response to that actions later in the change at least has access to the error message. We also need to add a note about this error to the docs, stating the response condition when a user denies location access. 
See https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API/Using_the_Geolocation_API
https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError

Gervwyk (approved):

PR #1551, opened 2023-04-06
Authentication fixes

SamTolmay:

This PR makes the following changes:
- A 200 API response is sent on head requests - this is required for email authentication on corporate networks: https://next-auth.js.org/tutorials/avoid-corporate-link-checking-email-provider
- The server-side session was not passed to the SessionProvider component, causing an extra network request to fetch the session on initial page load
- The window is now reloaded if a client was authenticated, and is no longer authenticated. This is to prevent the user looking at a stale page, and then failing to make requests on the page because they are unauthenticated.
- The basePath needed to be provided to the SessionProvider

## Checklist

- [X] Pull request is made to the "develop" branch
- [ ] Tests added
- [X] Code has been formatted with Prettier
- [X] Edits from maintainers are allowed

### DIFF
--- a/packages/server-dev/lib/auth/AuthConfigured.js
+++ b/packages/server-dev/lib/auth/AuthConfigured.js
@@ -15,14 +15,23 @@
 */
 /* eslint-disable react/jsx-props-no-spreading */
 
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import { SessionProvider, signIn, signOut, useSession } from 'next-auth/react';
 
 function Session({ children }) {
+  const wasAuthenticated = useRef(false);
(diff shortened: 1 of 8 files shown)


packages/server-dev/lib/auth/AuthConfigured.js
```diff
@@ -31,9 +42,12 @@ function Session({ children }) {
 
 function AuthConfigured({ authConfig, children, serverSession }) {
   const auth = { signIn, signOut, authConfig };
-
+  let basePath = process.env.LOWDEFY_BASE_PATH ?? lowdefyConfig.basePath;
```
Gervwyk, line 45: process.env?

PR #1596, opened 2023-05-31
fix(engine): Fix setting values of blocks to null using set state.

SamTolmay:

A bug in engine meant that if set state was called and the value for a block in state was undefined, the undefined value was not applied to the block, and the blocks previous value was retained. This PR fixes that bug.

## Checklist

- [x] Tests added

### DIFF
--- a/packages/engine/test/Block/setState.test.js
+++ b/packages/engine/test/Block/setState.test.js
@@ -94,7 +94,7 @@ test('Set block value to null using set state', async () => {
   expect(context.state).toEqual({ textInput: null });
 });
 
-test('v', async () => {
+test('Set nested block value to null using set state', async () => {
   const pageConfig = {
     id: 'root',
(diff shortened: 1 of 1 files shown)


packages/engine/test/Block/setState.test.js
```diff
@@ -0,0 +1,209 @@
...
+  const { button, textInput } = context._internal.RootBlocks.map;
+
+  textInput.setValue('abc');
+  expect(textInput.value).toBe('abc');
+  await button.triggerEvent({ name: 'onClick' });
+  expect(textInput.value).toBe(null);
+  expect(context.state).toEqual({ textInput: null });
+});
+
+test('v', async () => {
```
Gervwyk, line 97: Just a fix here.

Gervwyk (approved):

PR #1600, opened 2023-06-15
Improve logging

SamTolmay:
Closes #ISSUE_NUMBER

## Checklist

- [ ] Tests added


packages/server-dev/pages/api/page/[pageId].js
```diff
@@ -14,27 +14,19 @@
...
-    logger: console,
-    session,
-  });
+import apiWrapper from '../../../lib/server/apiWrapper.js';
 
+async function handler({ context, req, res }) {
   const { pageId } = req.query;
-  const pageConfig = await getPageConfig(apiContext, { pageId });
+  const pageConfig = await getPageConfig(context, { pageId });
+  // TODO: Do we log here?
```
Comment (Gervwyk) [inline packages/server-dev/pages/api/page/[pageId].js:24]: I think yes. We'd want to log the page id that was requested also. This can be useful for debugging role based error, where you've made a change which prohibits users from accessing a page because of their role, so both role, requested page id and the fact that 404 was served is important.

packages/server/lib/server/apiWrapper.js
```diff
@@ -0,0 +1,67 @@
...
+        logger,
+        operators,
+        secrets: getSecretsFromEnv(),
+        session,
+        req,
+      });
+      logRequest({ context });
+      // Await here so that if handler throws it is caught.
+      const response = await handler({ context, req, res });
+      // TODO: Log response time?
```
Gervwyk, line 55: Would be nice to have server response time available, for debugging log running requests. This is not something you want to only log on the client, since you'd want to exclude network time from this response time measurement.

Gervwyk (approved):

PR #1720, opened 2023-11-27
Text input type

Epictek:

Allows specify the field type of TextInput, personally I wanted this so I can use the native html datepicker for better mobile support but also should open up the possibilities for other behavior.

## Checklist

- [ ] Tests added


Gervwyk:
Hi @Epictek 
Thanks for the PR, would be happy to support this, I've tested with some of the options like `color` or `date` and getting some errors on `TextInput`, which should be simple to resolve. Will test / fix / evaluate further to include on next release if possible. 
In the meantime, you could also overwrite the `TextInput` block type in a plugin if this is blocking.

Gervwyk (approved):

Gervwyk (dismissed):

JohannMoller (approved):

packages/plugins/blocks/blocks-antd/src/blocks/TextInput/schema.json
```diff
@@ -9,6 +9,12 @@
         "default": false,
         "description": "Allow the user to clear their input."
       },
+      "type": {
+        "type": "string",
+        "enum": ["text", "number", "password", "tel", "email", "url"],
```
SamTolmay, line 14: We also have a `PasswordInput` block. Does it make sense to set type `password` on a normal text input? Same with number I guess?

Gervwyk: I think it does. Yes, there are purpose-built blocks, but this is a standard html implementation of <input type=".." /> which leaves it to the dev to decide how to use or not in edge cases like optimising for mobile etc.

Gervwyk (approved):

PR #1818, opened 2024-05-22
Add getRowId to AgGrid and AgGridInput to preserve selection.

JohannMoller:
~Closes #ISSUE_NUMBER~

This PR add the `getRowId` property to the AgGrid and AgGridInput blocks. This allows the selection on the table to be preserved even when the rowData changes (assuming the selected rows are in both data sets).
The row id is taken from the data. It first tries to use the field supplied in the `rowId` property from the data, if no `rowId` is supplied it tries the `_id` field, if there is no `_id` it using a JSON stringified string of the entire row as the rowId.

## Checklist

- [ ] Tests added

## Docs
Screenshot 2024-05-23 at 12 37 31

Screenshot 2024-05-23 at 12 38 29


Gervwyk (approved):

packages/plugins/blocks/blocks-aggrid/src/AgGridInput.js
```diff
@@ -29,6 +29,11 @@ const AgGridInput = ({ properties, methods, loading, events, value }) => {
 
   const memoDefaultColDef = useMemo(() => defaultColDef);
 
+  const getRowId = useCallback(
+    (params) => params.data[properties.rowId ?? '_id'] ?? JSON.stringify(params.data),
```
SamTolmay, line 33: An alternative implementation could be 
```js
params.data[properties.rowId] ?? params.data._id ?? params.data.id ?? JSON.stringify(params.data)
```
which also uses `id` since that is a more generic ID field than `_id` which is MongoDB specific?

JohannMoller: This makes sense. I updated this code in https://github.com/lowdefy/lowdefy/pull/1818/commits/51e45333b9b3ceff42bcbe3f94bcbb6710daf83f.

Gervwyk (approved):

PR #1865, opened 2024-11-05
Initialise routine handling

SamTolmay:


packages/api/src/routes/api/control/controlIf.test.js
```diff
@@ -0,0 +1,412 @@
...
+          },
+        },
+      },
+    ],
+  ]);
+  expect(res.response).toEqual(undefined);
+});
+
+// TODO: NOt catching err in test
+test.skip('missing :then', async () => {
```
Saiby100, line 407: Is this necessary since we do this at build?

PR #1869, opened 2024-11-05
Fixed merge conflicts and added return.

Saiby100:
Merge conflicts and added return control logic.


packages/api/src/routes/endpoints/control/controlIf.test.js
```diff
@@ -23,7 +23,7 @@ test.only('if condition is true', async () => {
     },
   };
   const { res, context } = await runTest({ routine });
-  expect(res.status).toEqual('continue');
+  expect(res.status).toBe('continue');
```
SamTolmay, line 26: These should be `toEqual`. `toBe` does not do string compare

SamTolmay (approved):

PR #1879, opened 2024-11-07
Add test context to runTest

JohannMoller:


packages/api/src/routes/endpoints/control/controlIf.test.js
```diff
@@ -41,17 +43,31 @@ test.only('if condition is true', async () => {
...
+          connectionId: 'testConnection',
+          requestId: 'requestId',
           properties: {
             response: 'Was true',
           },
         },
       },
     ],
+    [
+      {
```
SamTolmay, line 59: I think this should just be:
```
      {
        event: 'debug_end_request',
        requestResult: 'Was true',
   }    
```

SamTolmay (approved):

PR #1911, opened 2025-03-11
Fix aggrid input edit and row drag and drop.

Gervwyk:
PR #1818 broke editing aggrid cells and drag and drop functionality. Only applies when no row id is provided. 
This also removes the index property on change events. The rowIndex is the aggrid field that provides the index value

## Checklist

- [ ] Tests added

### DIFF
--- a/packages/plugins/blocks/blocks-aggrid/src/AgGridInput.js
+++ b/packages/plugins/blocks/blocks-aggrid/src/AgGridInput.js
@@ -195,8 +195,6 @@ const AgGridInput = ({ properties, methods, loading, events, value }) => {
     }
   }, [value]);
 
-  console.log(value);
-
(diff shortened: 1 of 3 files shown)


packages/plugins/blocks/blocks-aggrid/src/AgGridInput.js
```diff
@@ -198,6 +195,8 @@ const AgGridInput = ({ properties, methods, loading, events, value }) => {
     }
   }, [value]);
 
+  console.log(value);
```
JohannMoller, line 198: Here is a stray console.log

JohannMoller (dismissed):
Nice `assignRowId` function. There is just a console.log left over from testing

SamTolmay (dismissed):

SamTolmay (approved):

PR #1988, opened 2026-02-10
Update from main

Gervwyk:


packages/website/.eslintrc.json
```diff
@@ -0,0 +1,15 @@
+{
+  "root": true,
```
SamTolmay: We could probaly have a specific eslint config here if we have specific setting we want to set. But generally we use monorepo one

SamTolmay (approved):

PR #2124, opened 2026-04-16
Prevent white flash on page navigation in dark mode

StephanieJKS:
## Summary

When navigating between pages in dark mode, the screen would briefly flash white. This happens because `_app.js` uses `ssr: false`, so Ant Design's CSS-in-JS theme (which sets `--ant-color-bg-layout`) isn't available until hydration. The browser defaults to a white background in the gap between page load and theme application.

## Changes

- **@lowdefy/server, @lowdefy/server-dev**: Added a synchronous inline script in `_document.js` that reads the user's dark mode preference (config → localStorage → system `prefers-color-scheme`) and sets the `<html>` background to `#000` before first paint. Added `lib/build/theme.js` wrappers to import the `theme.json` build artifact for the `configDarkMode` value.
- **@lowdefy/build**: Added a `@layer base` CSS rule in the generated `globals.css` that sets `html.lowdefy { background-color: var(--ant-color-bg-layout) }`, so the background stays correct once Ant Design's CSS variables are active.
- **@lowdefy/client**: Added a mount-time `useEffect` in `useDarkMode` that removes the inline background style set by the `_document.js` script, allowing the CSS rule to take over (inline styles have higher specificity than CSS rules).

## Notes

The inline script in `_document.js` intentionally mirrors the same priority logic as `useDarkMode.js` (`configDarkMode` → `localStorage('lowdefy_darkMode')` → `prefers-color-scheme: dark`). If the dark mode resolution logic changes in `useDarkMode.js`, the inline script should be updated to match.


StephanieJKS:
## Documentation Updates

This PR triggered documentation improvements based on session insights.

### Files Updated
- `code-docs/architecture/theme-and-styling.md` — Added "Dark Mode Flash Prevention" section, updated globals.css structure, layer table, key files table, and debugging guide
- `code-docs/packages/client.md` — Added `useDarkMode.js` to Key Modules table

### Insights Captured
- The dark mode flash is caused by `_app.js` using `ssr: false`, creating a gap between page load and Ant Design CSS variable hydration
- Three coordinating pieces solve it: inline script (before paint), CSS rule (after hydration), mount cleanup (specificity handoff)
- The inline script must stay in sync with `useDarkMode.js` resolution logic

---
*Generated by `/l-docs-update`*

Gervwyk (approved):

PR #2143, opened 2026-04-29
fix(engine): Preserve input values across visible toggles via SetState

Gervwyk:
## Summary

When an input was inside a hidden container and a `SetState` action set its value (e.g. on `onInit`, before the container becomes visible), the value was silently reset to the input's `enforceType` default on the next `SetState` call. This made `SetState`-driven visibility toggles diverge from `setValue`-driven ones, which already preserved the value via `Block.evaluate`. This PR fixes the asymmetry in `Block.reset()` and adds tests covering inputs, lists, and the `_state` re-eval cascade.

## Changes

- **@lowdefy/engine**: `Block.reset()` now skips the `enforceType` fallback when the block was hidden in the previous eval cycle and has in-memory state to preserve — `this.value` for inputs, or existing `subSlots` for lists (which would otherwise be truncated by the rebuild loop). `Slots.updateState` continues to delete invisible blocks' state fields, and `Block.updateState` republishes the value on the visible→true transition, so the "invisible blocks have no state field" invariant is preserved across the gap.

## Breaking Changes

- Apps that relied on a hidden input being silently reset to its `enforceType` default by an unrelated `SetState` will now see the previously-set value preserved. To explicitly clear, use `SetState({ myInput: null })`.
- The user-facing `Reset` action (`createReset`) is unaffected for visible fields. Forms still submit only visible field values. List sub-slot rebuilding on `SetState({ list: [...] })` is unchanged (the explicit array still wins over preservation).

## Notes

- Twelve tests in `packages/engine/test/Block/container.test.js` cover: hidden→reveal preservation, `_state` cascade re-eval after the visibility flip (verifies the existing `Slots.updateStateFromRoot` repeat loop still propagates), `SetState` parity with `setValue` toggles, list sub-blocks across visibility cycles, the `Reset` action regression, explicit null clear (the documented mitigation), child input with its own `visible: false`, and explicit list replacement while hidden.
- The plan and a step-by-step trace of the bug live in the third commit's body and the included changeset.

## Test Exp Release: 0.0.0-experimental-20260429131524


Gervwyk:
0.0.0-experimental-20260429131524

SamTolmay (approved):

PR #2154, opened 2026-05-01
feat: Add i18n / locale support across antd, dayjs, and operators

Gervwyk:
Wires Ant Design's i18n (ConfigProvider locale + form.validateMessages)
and dayjs locale switching, and adds a translation framework (ICU
MessageFormat via intl-messageformat) so app authors can localize their
own strings via config and let users pick languages at runtime.

Schema: config.i18n with defaultLocale, fallbackLocale, locales (each
mapping a BCP 47 code to antd/dayjs locale names plus a label), and
messages keyed by locale.

Build: writeI18n.js emits i18n.json; codegenI18nLocales.js generates
build/i18n/antdLocales.js (dynamic-import map) and dayjsLocales.js so
only the locales the app declares are bundled.

Runtime: useLocale hook (mirrors useDarkMode) resolves active locale
from localStorage > navigator.languages > defaultLocale, exposes
window.__lowdefy_setLocale. _app.js threads antd locale +
validateMessages into ConfigProvider and runs dayjs.locale(). The
serverside API parses Accept-Language and threads i18n through
ServerParser. WebParser/ServerParser inject i18n as a first-class
field on the operator context.

Operators: _t (ICU MessageFormat key lookup with values + locale
override), _locale (active/default/fallback/supported view), both
client and server.

Action: SetLocale (mirrors SetDarkMode) writes preference and triggers
re-render.

Block: ConfigProvider antd block now accepts a locale prop for subtree
overrides.

https://claude.ai/code/session_01SBwAWkhRrjRtD71YQ5cXSC


Gervwyk:
Experimental version: 0.0.0-experimental-20260506095254

SamTolmay (approved):

PR #2157, opened 2026-05-05
docs(website): Add Lowdefy Agents article

machielvdw:
## Summary

Adds a release-style article introducing the agent system. Walks readers through what an agent is, the four-layer config model (connection, agent, tool, block), and how the integration leverages the Vercel AI SDK at a high level. Marked `draft: true` so it stays off the live articles list until reviewed and the publish date confirmed.

## Changes

- `packages/website/content/articles/lowdefy-agents.md` — the article (~1700 words). Frontmatter follows the existing convention (title, subtitle, authorId, publishedAt, readTimeMinutes, tags, draft).
- `packages/website/src/content/articles/index.js` — import + register the article in the array.

## Article shape

- Opening: state the result (streaming AI conversation in YAML), name the unique angle (existing API endpoints become AI tools).
- **What an agent is** — plain-language definition of an LLM in a tool loop, then the four-layer mapping table.
- **Adding an agent** — five progressively richer YAML examples: a 30-line chatbot, endpoints as tools (with `InternalApi`), MCP servers as connections, tool confirmation (`confirm: true`), sub-agents.
- **Under the hood** — high-level technical breakdown of how the four config layers compose onto the Vercel AI SDK runtime.

## Test plan

- [ ] Build the website locally and verify the article renders at `/articles/lowdefy-agents` when `draft` is flipped to `false`.
- [ ] Confirm the article does not appear on the index list while `draft: true`.
- [ ] Spot-check the YAML examples against the agent test app patterns.
- [ ] Sam/Gerrie review for tone, accuracy, and final publish date.


Gervwyk:
@machielvdw lets add a nice gif of the chat and a small example + gif of the set state in action?

packages/website/content/articles/lowdefy-agents.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Adding an agent
+
+### A chatbot in around thirty lines
+
+```yaml
+plugins:
+  - name: "@lowdefy/connection-anthropic"
+    version: 5.3.0
+  - name: "@lowdefy/blocks-antd-x"
```
JohannMoller: Is this plugin necessary?

machielvdw:
> @machielvdw lets add a nice gif of the chat and a small example + gif of the set state in action?

Added in 47a2ce64aaa88bba6104d4455984ab11c7b7ca0c

packages/website/content/articles/lowdefy-agents.md
machielvdw: You don't have to explicitly add it. Will remove it, thanks.

Also need to update the version number to what the new lowdefy version will be after release for `@lowdefy/connection-anthropic`

SamTolmay:
Very nice.

I think title should be "Lowdefy v5.3: {Some Title}" format (or whatever version it is) to be consistant with other release announcments.

I also think the article might be a little to technical - I don't think users care to much exactly what happens under the hood - more about what they can do with it, and that it "just works". I think there is some value in saying "we are doing lots of stuff under the hood", but think this might be a little too technical in some places

machielvdw:
> Very nice.
> 
> I think title should be "Lowdefy v5.3: {Some Title}" format (or whatever version it is) to be consistant with other release announcments.
> 
> I also think the article might be a little to technical - I don't think users care to much exactly what happens under the hood - more about what they can do with it, and that it "just works". I think there is some value in saying "we are doing lots of stuff under the hood", but think this might be a little too technical in some places

Cool. I wanted to structure the article with a short brief at the beginning about what AI agents are, then going into what you can do in Lowdefy with AI agents, and then a high level technical breakdown. I agree with you that the technical breakdown got a bit too in depth. I'll either take it out or maybe replace it with a section showing some more cool features of Lowdefy agents.

PR #2172, opened 2026-05-13
fix: Unblock Playwright e2e for v5+ Lowdefy apps and add list-block helpers

machielvdw:
## Summary

Four things were blocking Playwright e2e tests for v5+ Lowdefy apps:

- `@lowdefy/server-e2e` had drifted behind `@lowdefy/server` — Turbopack incompat on Next 16, single-segment routes instead of catch-alls, stale dark-mode rendering, several build-time and plugin-loading bugs.
- `@lowdefy/e2e-utils` `extractBlockMap` didn't walk the `slots` container shape, which is the standard compiled shape in v5+, so `ldf.block('<nested-id>')` failed on every non-root block for every v5+ app.
- The `lowdefy` CLI silently reused whichever server package landed in `.lowdefy/server/` first, ignoring later `--server` flag changes.
- `ldf.block(...)` had no support for list-iterated children — every per-row toggle/input/button forced raw `page.locator('[id="..."]')` selectors that coupled tests to per-block DOM internals (e.g. Switch's `_input` suffix) and lost the common helper assertions (`expect.checked`, `.visible`, `.validationError`).

This PR resyncs and fixes all of them so `lowdefy build --server e2e` produces a working app, `ldf.block(...)` resolves nested blocks (including list-iterated children), and `ldf.list(listId)` provides ergonomic row addressing.

```js
// Before: raw page.locator, brittle, no helper assertions
await ldf.page.locator('[id="legal_rows.0.toggle_input"]').click();

// After: first-class block helper, by index or by key
await ldf.block('legal_rows.0.toggle').do.toggle();
const row = await ldf.list('legal_rows').rowBy('_id', 'bbbee');
await row.block('toggle').do.toggle();
await row.block('toggle').expect.checked();
```

### `@lowdefy/server-e2e`

- **Turbopack / next.config.js** — drop the legacy webpack `resolve.fallback` polyfill block, add `turbopack: {}`, switch `transpilePackages` to mirror server (`@lowdefy/client`, `@ant-design/x`, `@ant-design/x-markdown`, `...blockPackages` from `build/blockPackages.json`).
- **Catch-all routes** — `pages/[[...pageId]].js`, `pages/api/endpoints/[...endpointId].js`, `pages/api/request/[...path].js`. Apps with nested page paths (e.g. `pages: [{ id: 'foo/bar' }]`) now render instead of 404.
- **Dark-mode rendering parity** — `_app.js` uses `useDarkMode` from `@lowdefy/client` and wraps children in `ThemeTokenResolver` (publishes the resolved antd token onto `lowdefy.theme._resolvedAntdToken`). `_document.js` re-adds the pre-hydration background-color script that prevents the dark/light flash on navigation. New `lib/build/theme.js` wrapper around `build/theme.json`.
- **Client-error hardening** — `pages/api/client-error.js` enforces the same-origin host check (403 on missing origin or host mismatch) and strips `req.body['~e'].received` before forwarding to `logClientError`.
- **build.mjs + missing-yaml guard** — switch `build.mjs` to `error instanceof BuildError` (drop the obsolete `error.isFormatted || error.hideStack` shape and the unused `mixin` logger config); add the `if (!lowdefyYaml) return [];` guard in `getPluginDefinitions`. Covered by a new unit test.
- **Plugin loading** — `createCustomPluginTypesMap` unwraps the ESM `default` export from `require('<plugin>/types')` and initializes `blockMetas: {}` so plugin block metadata is collected. Jest config + first unit tests added for the package.
- **Dependencies** — add `@lowdefy/blocks-antd-x`; move `tailwindcss` and `@tailwindcss/postcss` into `dependencies`; remove the now-unused `process` browser polyfill dep.

### `@lowdefy/e2e-utils`

- **`extractBlockMap` slots traversal** — walk `slots.<name>.blocks` alongside `areas.<name>.blocks` and `blocks`. Compiled page artifacts under `.lowdefy/server/build/pages/<pageId>.json` use the `slots` container shape in v5+, which the walker was skipping — so `generateManifest` only saw the page root, and `ldf.block('<any-nested-id>')` threw `Block "<id>" not found on page. Available blocks: <pageId>` for every non-root block. New co-located test file covers the slots shape, mixed slots+areas, multiple slot names, the legacy areas shape, and unregistered-type skip.
- **List-child blocks in `ldf.block()`** — build emits templated keys (`legal_rows.$.toggle`) for `category: 'list'` children by carrying a path prefix through `extractBlockMap` and injecting `{listId}.$.` when entering a list block. `generateManifest` now reads `plugins/blockMetas.json` alongside `types.json` (missing file falls back gracefully). Runtime resolves numeric segments in literal blockIds back to `$` via a new pure `resolveTemplateId` helper, so individual block `e2e.js` files needed no changes. Nested lists stack prefixes (`outer.$.inner.$.btn`).
- **New `ldf.list(listId)` helper** — `.count()`, `.row(i)`, `.rowBy(key, value)`, `.rowWhere(predicate)`. `row(i)` is synchronous; `rowBy`/`rowWhere` are async (state read). `rowBy` deliberately takes `(key, value)` with no `_id` default — the name never lies even when matching on `slug`, `type`, etc.

### `lowdefy` CLI

- **`getServer` re-fetches on package-name mismatch** — both `lowdefy build` and `lowdefy build --server e2e` write to `.lowdefy/server/`, so the previous version-only cache check let flipping between them silently reuse whichever package landed there first. Now we compare both name and version. Adds `getServer.test.js` covering the local-version skip, missing-directory, matching, version-mismatch, and name-mismatch paths.

## Intentionally unchanged (not drift, deliberate trim-downs)

- **Sentry** — no `sentry.*.config.js`, no `lib/{client,server}/sentry/`, no init in `_app.js`, no capture in `client-error.js`, no `setSentryUser` in `apiWrapper.js`.
- **NextAuth** — no `pages/api/auth/[...nextauth].js`. Auth is the cookie-mock implementation in `lib/client/auth/Auth.js` + `pages/api/auth/session.js`.
- **Agents** — no `pages/api/agent/[...path].js`, no `agents` on the apiWrapper context, no `outputFileTracingIncludes` in `next.config.js`.
- **Vercel install scripts** (`vercel.install.*.sh`).
- **Row-scoped state/request helpers** (`ldf.list().row(i).state(field)`) — narrower hit-rate since users frequently maintain a separate state map keyed by some domain id. Keep using ``ldf.state(`${listId}.${i}.field`)`` for now.

## Test plan

- [x] `pnpm --filter=@lowdefy/server-e2e test` — 4 tests pass (ESM unwrap, blockMetas, icons, missing-yaml)
- [x] `pnpm --filter=@lowdefy/e2e-utils test` — passes (incl. slots-traversal, list-child, and `resolveTemplateId` cases)
- [x] `pnpm --filter=lowdefy test` — 56 tests pass (CLI), incl. 5 new getServer cases
- [x] `pnpm build` — full repo compiles
- [x] Manual: `lowdefy build --server e2e` against any v5+ app prints no Turbopack error
- [x] Manual: `ldf.block('welcome_card').expect.visible()` succeeds for non-root blocks instead of throwing `Available blocks: <pageId>`
- [x] Manual: `ldf.block('legal_rows.0.toggle')` and `ldf.list('legal_rows').rowBy('_id', 'x').block('toggle')` resolve to the correct iterated child
- [ ] Manual: app with nested page paths (`pages: [{ id: 'foo/bar' }]`) renders at `/foo/bar` instead of 404
- [ ] Manual: dark-themed app under e2e does not flash light before hydration
- [ ] Manual: app with no `lowdefy.yaml` still builds under `--server e2e`


machielvdw:
0.0.0-experimental-20260514071610

machielvdw:
0.0.0-experimental-20260514080801

machielvdw:
0.0.0-experimental-20260514082957

SamTolmay (approved):

PR #2188, opened 2026-06-02
Fix false CallAPI non-existent-endpoint warnings in dev server (top-level + module endpoints)

machielvdw:
## Summary

Under `lowdefy dev`, every `CallAPI` action logged a false `[ConfigWarning] ... references non-existent endpoint "..."`, even when the endpoint was correctly defined under `api:` and worked at runtime. A full `lowdefy build` was unaffected.

The JIT page builder validates CallAPI references against `context.components.api`, but the dev server's `getBuildContext` rebuilds its context from disk artifacts. This PR restores the api endpoints from the persisted `build/api/*.json` artifacts so the dev-time checks (`validateCallApiRefs`) match a full build.

There were **two layers** to the bug:

1. **Top-level endpoints** — `getBuildContext` never restored `components.api`, so the endpoint set was always empty and every CallAPI was flagged.
2. **Module endpoints** — module endpoint ids are scoped with the module entry id (`<moduleId>/<endpointId>`), so `writeApi` writes them to a _nested_ path (`build/api/<moduleId>/<endpointId>.json`). The first version of the reader did a flat `readdirSync` and skipped subdirectories, so every module CallAPI was still flagged. The reader now walks the api directory recursively.

## Changes

- **@lowdefy/server-dev**
  - New `readBuildApiArtifacts` helper **recursively** reads and deserializes `build/api/**/*.json`; `getBuildContext` sets `context.components = { api: ... }` from it.
  - `validateCallApiRefs`' `non-existent endpoint` and `InternalApi`-not-client-callable checks now behave the same in dev as in a full build, for **both top-level and module endpoints**.
  - Returns `[]` when no endpoints exist (matching `components.api = []`), so a CallAPI to a genuinely missing endpoint still correctly warns.
- **@lowdefy/build** (test-only)
  - New `buildPageJitCallApi.integration.test.js`: a full-pipeline test (`shallowBuild` → real `build/api/*.json` on disk → `buildPageJit`) covering a top-level endpoint, a module endpoint resolved via `_module.endpointId`, and a genuinely-missing endpoint. This crosses the seam the existing unit test bypassed (it set `context.components.api` directly, so it could never catch the nested-path read).

## Notes

- **Fix at the source:** the nested `build/api/<moduleId>/...` layout is correct and is what the runtime executor reads by path; only the dev hydration's flat directory scan was wrong, so the reader was made recursive rather than changing the write path.
- No extra cache-invalidation work was needed: api endpoint files are skeleton sources, so editing one re-runs `shallowBuild`, which rewrites both `build/api/*.json` and `pageRegistry.json`; the new `pageRegistry.json` mtime resets the cached dev context.
- Patch changeset included for `@lowdefy/server-dev`.

## Follow-ups (pre-existing, not addressed here)

These predate this branch (in `develop` already) and are unverified code-reading observations:

- `buildPageJit` doesn't reset `context.callApiActionRefs` between page builds on the shared cached dev context, so refs accumulate across pages in a session (can mis-attribute/duplicate other CallAPI warnings; won't cause "non-existent" warnings).
- On the cached path, `buildPageIfNeeded` returns bare `true`, so `pageConfig._warnings` is never re-attached — warnings only show on a page's first build.

### DIFF
--- a/.changeset/fix-jit-callapi-endpoint-refs.md
+++ b/.changeset/fix-jit-callapi-endpoint-refs.md
@@ -1,0 +1,9 @@
+---
+'@lowdefy/server-dev': patch
+---
+
+fix(server-dev): Stop false "non-existent endpoint" warnings for CallAPI actions in dev.
+
+Under `lowdefy dev`, every `CallAPI` action logged a `ConfigWarning` that its endpoint did not exist, even when the endpoint was correctly defined under `api:` and worked at runtime. A full `lowdefy build` was unaffected.
+
+The JIT page builder validates CallAPI references against `context.components.api`, but the dev server's build context (`getBuildContext`) is rebuilt from disk artifacts and never restored the api endpoints, so the endpoint set was always empty and every reference was flagged. The dev context now hydrates the endpoints from the persisted `build/api/<endpointId>.json` artifacts (written by the skeleton build), so the non-existent-endpoint and InternalApi-not-client-callable checks behave the same in dev as in a full build.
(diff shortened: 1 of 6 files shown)


Gervwyk:
included in 0.0.0-experimental-20260602160131 but does not resolve imo, please check.

machielvdw:
> included in 0.0.0-experimental-20260602160131 but does not resolve imo, please check.

Thanks for checking @Gervwyk. I'll have a look.

machielvdw:
@Gervwyk I cut a new release based off this branch `0.0.0-experimental-20260603082728` (might not have everything you need from other open PRs). Made some changes and it seems to be working now.

PR #2197, opened 2026-06-18
Agent fixes: full onFinish messages, auto conversationId, generateTitle

machielvdw:
## Summary

The agents feature is new, and the first production app to use it surfaced three gaps plus a documentation hole. This PR fixes a server-side persistence correctness bug, removes a race around conversation ids, adds built-in title generation, and documents the onFinish payload contract and the new options.

## Changes

- **@lowdefy/ai-utils**:
  - **Fix (#1):** the agent `onFinish` hook payload `messages` previously contained only the request input, so a `save-conversation` hook persisted the user turn but never the assistant reply. `handleAgentChat` now captures the final UI message list (input + generated assistant message, including tool parts) from the stream's `onFinish` on both the default and prune paths, and uses it in the payload (`capturedMessages ?? messages`, falling back to input only on error/abort).
  - **Feat (#4):** new `generateTitle` agent option (`boolean | { model }`). On the first turn it generates a short title from the first user message via a one-shot `generateText`, run concurrently with the response, and emits a `data-chat-title` part that fires `onTitleGenerated`. Off by default; failures are non-fatal.
- **@lowdefy/blocks-antd-x**:
  - **Feat (#3):** `AgentChat` auto-mints a stable `conversationId` at mount when the prop is empty, so a turn sent before the app assigns an id (e.g. a welcome prompt) no longer posts an undefined id. The effective id is surfaced once per conversation, on its first user message, via a new `onConversationStart` event. App-supplied ids remain authoritative. Flows through the existing URL transport — no API/route changes.
- **@lowdefy/docs** (agents):
  - Documented the onFinish payload contract (added `usage`, clarified `messages` includes the assistant reply and that context fields are top-level), the `generateTitle` property, and the `onConversationStart` event / conversationId minting behavior.

## Notes

- No `@lowdefy/api` or server-route changes were needed — `conversationId` already flows via the request URL.
- The #1 fix and #4 feature share `handleAgentChat.js`/its test, so they are in one ai-utils commit; the two changesets preserve the patch (fix) vs minor (feat) bump distinction.
- Verified: `@lowdefy/ai-utils` tests 124/124 pass; both changed packages build clean.


packages/plugins/blocks/blocks-antd-x/src/blocks/AgentChat/AgentChat.js
```diff
@@ -68,11 +68,27 @@ function AgentChat({ blockId, components: { Icon }, methods, pageId, properties
     return initial;
   });
 
+  // When no conversationId is supplied, mint a stable session id so every turn
+  // (including a welcome prompt sent before the app sets the prop) posts a
+  // consistent id. App-supplied ids remain authoritative.
+  const mintedIdRef = useRef(null);
+  if (!conversationId && !mintedIdRef.current) {
+    mintedIdRef.current =
+      crypto.randomUUID?.() ?? `c_${Date.now()}_${Math.random().toString(36).slice(2)}`;
```
machielvdw: Why the fallback?

machielvdw: Good catch — dropped it. The fallback only guarded plain-HTTP non-localhost contexts where `crypto.randomUUID` is `undefined` (it requires a secure context). Lowdefy apps run on HTTPS or localhost, so it guarded a case that doesn't happen in practice. Now just `crypto.randomUUID()` in a2c5ac75a.

Gervwyk (approved):

PR #2211, opened 2026-07-01
Scheduled API endpoints (cron) via Vercel Build Output API

machielvdw:
## Summary

Adds a `schedules` key to `Api`/`InternalApi` endpoints so a routine can run on a cron timer, and delivers it on Vercel via the [Build Output API](https://vercel.com/docs/build-output-api). Because `vercel.json` is read *before* the build, build-generated crons could only land in it by committing them; the Build Output API `config.json` is read *after* the build and has a first-class `crons` field, so the cron jobs are generated from the declared schedules on every deploy with nothing committed by hand.

```yaml
id: purge-stale-conversations
type: Api
schedules:
  - cron: '0 6 * * *'
    payload: { mode: full }
  - cron: '*/15 * * * *'
    payload: { mode: incremental }
routine:
  - id: purge
    type: MongoDBDeleteMany
    connectionId: conversations
    properties:
      filter: { updatedAt: { $lt: { _payload: cutoff } } }
```

## Changes

- **@lowdefy/build**: `schedules` added to the endpoint schema; validated against Vercel's restricted cron subset (5 fields, no `MON`/`JAN`, DOM/DOW mutually exclusive) with unique crons per endpoint; passed through to the endpoint artifact and flattened into a `build/schedules.json` manifest.
- **@lowdefy/api**: new `runScheduledEndpoint` runs the routine as a **system context** (no user session — `_user` is `undefined`), resolving the payload from the firing schedule; does not block `InternalApi` and does not consult the endpoint's `auth`.
- **@lowdefy/server + @lowdefy/server-dev**: new `GET /api/cron/*` route, secured by the `CRON_SECRET` env var (fails closed), reading the `x-vercel-cron-schedule` header to select the schedule.
- **@lowdefy/cli**: Vercel deploy migrated from `vercel.json` `functions`+`rewrites` to the Build Output API. New `lowdefy vercel-output` command assembles `.vercel/output/` (static + one `api.func` + `config.json`) and generates the `crons` array from `schedules.json`. `init-vercel` now scaffolds `vercel.build.sh`; the function handler is generated at build time rather than committed.
- **docs**: Build Output API deployment topology + a "Scheduled endpoints (cron)" section (`packages/docs/deployment/vercel.yaml`), and the `schedules` key on the API concepts page.

## Key Files

- `packages/api/src/routes/endpoints/runScheduledEndpoint.js` — schedule matching + system-context routine execution
- `packages/servers/server/src/routes/cron.js` — `/api/cron/*` handler, `CRON_SECRET` gate
- `packages/build/src/utils/validateCronExpression.js` — Vercel cron-subset validator (build-time)
- `packages/cli/src/commands/vercelOutput/vercelOutput.js` — `.vercel/output` assembly + crons generation
- `packages/build/src/build/writeApi.js` — `build/schedules.json` manifest

## Breaking Changes

- This migrates how **all** Lowdefy Vercel apps deploy (not only scheduled ones). After upgrading, re-run `lowdefy init-vercel` (or update the `deploy/` files by hand) — the scaffold now uses `vercel.build.sh` and the Build Output API instead of `vercel.json` `functions`/`rewrites`.

## Notes

- Cron jobs run in **UTC** and are subject to Vercel plan limits: **Hobby** allows daily only (a sub-daily expression fails the deployment), **Pro**/**Enterprise** allow per-minute; 100 crons/project max. Delivery is best-effort and not retried, so scheduled routines should be idempotent.
- **Verified**: unit tests (build + api), the new snapshot fixture, an assembly smoke test of `.vercel/output`, full `pnpm build`, prettier + eslint.
- **Not yet verified on a real deployment** — needs confirming on a Vercel Pro project that (a) the crons appear/fire and (b) the single-function catch-all route preserves the original `req.url` to the Hono app. This is the one load-bearing assumption of the Build Output API migration.


packages/api/src/routes/endpoints/runScheduledEndpoint.js
```diff
@@ -0,0 +1,85 @@
...
+  if (cron) {
+    schedule = schedules.find((s) => s.cron === cron);
+  } else if (schedules.length === 1) {
+    schedule = schedules[0];
+  }
+  if (!schedule) {
+    throw new ConfigError(`No schedule matching cron "${cron}" for API Endpoint "${endpointId}".`);
+  }
+
+  // Force a system context regardless of any session cookie sent with the request.
```
machielvdw, line 57: Is this what we want?

machielvdw:
0.0.0-experimental-20260701143405

machielvdw:
0.0.0-experimental-20260702084155

PR #2216, opened 2026-07-02
feat(auth): Phase 2 - Hooks (user tier)

SamTolmay:
## Summary

Phase 2 of the auth upgrade: the user-tier hook system. Each `auth.hooks` entry binds an auth lifecycle point to an `InternalApi` endpoint whose routine is the hook body, invoked in a fresh system context with full Lowdefy access (connections, operators, control flow). Implements the binding model, the system-context bridge, the payload catalog, the return convention, and slot composition per [hooks/design.md](https://github.com/lowdefy/lowdefy-design/blob/main/designs/auth-upgrade/hooks/design.md); the `auth.hooks` surface per config-schema/design.md.

Stacked on #2215 (phase 1) — retargets to `auth-upgrade` when it merges.

## Changes

- **@lowdefy/build**: `hooks` registered in the `authConfig` schema (closed schema, `{ id, point, endpointId }`); new `buildAuthHooks` step validates known point (frozen 13-point catalog incl. `invitation.send`), endpointId resolves to an existing `type: InternalApi` endpoint, and one hook per point; `setAuthDefaults` writes `hooks: []`; `auth.json` carries the array. No retired callbacks/events discovery existed in build — nothing to remove there.
- **@lowdefy/api**: `createSystemContext` builds a fresh off-request context per hook fire (empty `user`, authorize bypassed — the auth engine is a trusted internal caller and `InternalApi` blocks HTTP). `getBetterAuthConfig` closes over the factory and assembles `databaseHooks` plus the `email.verified` → `afterEmailVerification` synthetic callback from `authJson.hooks`. One composed slot per point, engine bindings first (registry empty until phase 3), then the user hook. The dispatch adapter maps point data to the catalog payload, calls the shipped `invokeEndpoint` unchanged (depth-limited), and translates the routine envelope per the return convention: `:return` data wrapped as `{ data }` for database `before`, `:reject` → `APIError` carrying the reject message, throw → rethrow, returns ignored on `after`/synthetic. Each fire emits a `debug_auth_hook` log with `elapsed_ms`.
- **@lowdefy/server / server-dev**: each server assembles the `createSystemContext` factory over its own singletons and passes it into `getBetterAuth`; server-dev re-reads the dynamic jsMap per fire (loader extracted from the middleware). The retired `callbacks`/`events` keys removed from all `createCustomPluginTypesMap` templates (server, server-dev, server-e2e).
- **apps/auth-reference**: hook scenarios 11–16 in the README; `InternalApi` endpoints bound to `user.create.before` (record replace + domain veto), `session.create.after` (audit row via a MongoDB connection, proving payload shape / empty `_user` / `_secret`), `email.verified`; a protected `hook-audit` page lists the trail.

## Key Files

- `packages/api/src/context/createSystemContext.js` — the system-context bridge
- `packages/api/src/routes/auth/hooks/buildHooks.js` — slot assembly from engine + user bindings
- `packages/api/src/routes/auth/hooks/createUserBeforeHook.js` — return-convention translation
- `packages/api/src/routes/auth/hooks/authHookPoints.js` — runtime payload catalog
- `packages/build/src/build/buildAuth/buildAuthHooks.js` — build validation

## Gate

Verified two ways: unit tests (`pnpm test` green: build 1539, api 466, full suite exit 0) and an end-to-end harness driving the reference app's real build artifacts through a real BetterAuth 1.6.23 instance against MongoDB (memory server) and a live SMTP sink — 10/10 checks passed.

- ✅ Reference app binds `InternalApi` endpoints at a database `before` point (`user.create.before`), a database `after` point (`session.create.after`), and `email.verified`; each receives the catalog's payload shape (asserted via `payloadKeys` stored in the audit rows: `['user']`, `['session','user']`).
- ✅ A `user.create.before` hook's `:return` replaces the written record (name uppercased in the stored `users` row).
- ✅ A `:reject` in a `before` hook vetoes end to end — signup fails with the reject message, zero rows written.
- ✅ An `after` hook fires without affecting the committed write; a deliberate throw surfaces as an operational error while the session row stands (`:try` guidance documented in the app and README).
- ✅ Inside a hook routine `_user` is empty, `_payload` holds the subject, `_secret` resolves (asserted as stored booleans in the audit row).
- ✅ Build rejects: unknown `point`, missing `endpointId`, `endpointId` of `type: Api`, and two entries binding the same point — each verified against the real app build with clear `ConfigError`s.
- ✅ Unit tests for the dispatch adapter and return-convention translation (`buildHooks.test.js`, `createUserBeforeHook.test.js`, `composeBeforeSlot.test.js`, `authHookPoints.test.js`, `createSystemContext.test.js`); full suite passes.

## Report back — session.create hook latency

Measured on the e2e harness (M-series mac, local memory-server MongoDB):

- **No-op hook fire on `session.create` (context build + payload user fetch + `invokeEndpoint`): mean 0.25 ms, p50 0.23 ms, p95 0.32 ms (n=30).**
- `createSystemContext()` alone: ~0.3 µs per call.
- Sign-in wall time 40.83 ms with the hook vs 40.72 ms without (password hashing dominates) — the added cost is ~0.25 ms, ~0.6% of a login. No optimization warranted.

## Design learnings / escalations

1. **`user.update` payload deviation (catalog amendment recommended).** BetterAuth 1.6.23 hands `update.before` only the changed fields (no subject id — `updateWithHooks` passes the raw `data`) and `update.after` only the updated record. The catalog's `{ user, changes }` cannot be filled on both sides; shipped as `{ user: null, changes }` (before) and `{ user, changes: null }` (after). Recommend the payload catalog document exactly this. (Session/account points *do* get the full `user` — fetched via `internalAdapter.findUserById`, one extra read per fire, included in the latency number.)
2. **`:reject` maps to `APIError('BAD_REQUEST', { message })`** rather than bare `false`, per the design's "(or an `APIError`)" allowance — the routine's reject message surfaces inline to the client instead of BetterAuth's generic `FAILED_TO_CREATE_USER`.
3. **After-hook throws** propagate out of `auth.handler()` as a rejected promise (BetterAuth awaits queued after-transaction hooks inside the request chain), which Hono's error middleware turns into a 500 — the "operational error" of the convention. Confirmed against the installed source (`@better-auth/core` `runWithAdapter` awaits `pendingHooks` before returning).
4. `invitation.send` is accepted at build per the frozen point set; at runtime a startup warning states the binding will not fire until the org plugin lands (phase 3/6).
5. `code-docs/architecture/auth-system.md` predates phase 1 and still describes the NextAuth build; not updated here — flagging for a docs pass once the auth phases settle.

### DIFF
--- a/apps/auth-reference/README.md
+++ b/apps/auth-reference/README.md
@@ -3,7 +3,9 @@
 The living definition of done for the auth upgrade. Phase 1 exercises the
 BetterAuth engine: email/password signup with email verification, magic
 link, OAuth, protected/public/role-gated pages, sign out, and session
-revocation. Each later phase grows this app with a scenario.
+revocation. Phase 2 adds auth hooks: `InternalApi` endpoints bound to
+`user.create.before`, `session.create.after`, and `email.verified`. Each
+later phase grows this app with a scenario.
 
 ## Prerequisites
 
@@ -29,6 +31,8 @@ revocation. Each later phase grows this app with a scenario.
   LOWDEFY_SECRET_AUTH_DATABASE_URI=mongodb://localhost:27017/auth-reference
   LOWDEFY_SECRET_BETTER_AUTH_SECRET=<openssl rand -base64 32>
   LOWDEFY_SECRET_SMTP_HOST=localhost
+  # The hook scenarios assert this exact value lands in the audit row:
+  LOWDEFY_SECRET_HOOK_AUDIT_KEY=audit-secret-value
   # Only needed to exercise the OAuth scenario:
   LOWDEFY_SECRET_GOOGLE_CLIENT_ID=<from Google Cloud console>
   LOWDEFY_SECRET_GOOGLE_CLIENT_SECRET=<from Google Cloud console>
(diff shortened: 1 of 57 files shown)


SamTolmay:

PR #2217, opened 2026-07-02
feat(auth): Phase 3 - user model and organizations

SamTolmay:
Phase 3 of the auth upgrade — the user model and organizations. Implements [phase-3-user-model.md](https://github.com/lowdefy/lowdefy-design/blob/main/designs/auth-upgrade/tasks/phase-3-user-model.md) against the designs in `designs/auth-upgrade/` (user-model, hooks, mongodb, config-schema, admin Decision 5).

> **Stacking note:** phases 1–2 are not yet merged into `auth-upgrade`, so this PR stacks on `auth-upgrade-phase-2-hooks` (which contains phases 1, 1-a and 2). GitHub will re-target it to `auth-upgrade` when that branch merges.

## What changed per package

**@lowdefy/build**
- `auth.organizations` registered in the schema (`policy: pinned | tenant`, `org` slug, `signup: invite-only | open`), `additionalProperties: false`, with coherence validation: an explicitly pinned deployment must name its `org`; `org` is rejected under `tenant`; `signup` is ignored under `tenant` per the design. An omitted block defaults to the auto-seeded `default` org, pinned, invite-only.
- `buildRoleCatalog` collects every role name from `auth.pages.roles` / `auth.api.roles` / `auth.websockets.roles` into `auth.roles` — the catalog the runtime registers in the org plugin's access control with empty statements. Role names containing commas fail the build (CSV storage on `member.role`).
- All build fixture snapshots regenerated (every `auth.json` now carries `roles` and, when configured, `organizations` defaults).

**@lowdefy/api**
- The Organization plugin is always on: catalog roles registered in its AC with empty statements; built-in `owner`/`admin`/`member` keep their plugin statements (BetterAuth's permission check *replaces* rather than merges defaults when `roles` is passed — re-included explicitly, and a catalog role reusing a built-in name does not clobber it); `allowUserToCreateOrganization: false` (org creation is engine-only; system actions bypass the switch); org models mapped to `user-organizations` / `user-members` / `user-invitations`.
- Internal additionalFields: `contactId` (string) + `attributes` (json) on `user`, `attributes` (json) on `member`, `contactId` (string) on `invitation`. Not an app-facing surface — there is no `auth.user` key.
- Engine-tier hooks fill phase 2's slots (`engineHooks.js` placeholder → `buildEngineHooks` factory):
  - `session.create.before` applies the active-org policy. **pinned**: member → active org set; non-member with a pending, unexpired invitation → session admitted with no active org (the wall still gates every page until accept); otherwise the session is aborted with a 403 `APIError`, code `MEMBERSHIP_REQUIRED`, surfaced inline by the client sign-in call. **tenant**: oldest membership wins; pending invitation → session, mint nothing; otherwise the user's own organization is minted lazily (owner, active).
  - `user.create.after` auto-joins the pinned org with the built-in `member` role on open signup (idempotent — see learnings).
  - `organizationHooks.afterAcceptInvitation` stamps `invitation.contactId` onto the accepting user.
  - `invitation.send` is now wired to the org plugin's `sendInvitationEmail`; a bound hook owns the email, otherwise a stock template goes through `auth.email`, otherwise sending fails with an error naming the fix.
- Startup ensure-by-slug seeding of the pinned org (created if missing, memoized, unique-slug-index backstopped).
- `resolveAuthentication`: the live member read by `{ userId, activeOrganizationId }` on every authenticated request — one indexed lookup serving the hard wall (no row ⇒ unauthenticated, never logged-in-with-no-roles) and role resolution (`member.role` CSV split). `context.user.attributes` is the shallow merge of `user.attributes` and the active member's `attributes`, member wins. Session branch stays terminal.

**@lowdefy/client, @lowdefy/engine, @lowdefy/actions-core, servers**
- `SetActiveOrganization` action (client method → `organization.setActive` via the org client plugin), chained with `UpdateSession` for re-sync.
- Roles/attributes are not carried by the base session, so the servers expose the resolved caller at `GET /api/user` (same `resolveAuthentication` output the first render embeds); `UpdateSession` now re-syncs from it instead of pinning `roles: []` (the phase-1 interim contract), and the `Session` component keeps the last server-resolved caller authoritative while the session user is unchanged.

**apps/auth-reference (+ two new sibling apps)**
- `auth-reference` pins `org-a`, open signup: merge-on-signup endpoints (`user.create.before` inline match for verified OAuth; `email.verified` explicit update for email/password), contacts page (string `_id`s), members admin page driving the plugin's own `inviteMember`/`updateMemberRole` from an owner/admin session, public accept-invitation page, organizations switcher, auditor role-gated page, inline `MEMBERSHIP_REQUIRED` login UX.
- `auth-reference-b` pins `org-b`, invite-only, same auth database — the hard wall between deployments and the invitation flow.
- `auth-reference-tenant` runs the tenant policy on its own database.
- `scripts/set-member.mjs` — walkthrough bootstrap (same standing as the migration codemod) for promoting roles/attributes and removing member rows; README scenarios 17–28 cover the phase-3 gate.

## Gate, item by item

Verified three ways: unit tests (in-repo), a runtime probe driving the real `getBetterAuth` assembly against `better-auth@1.6.23` on an in-memory MongoDB (21/21 checks), and the reference-app walkthrough (manual, documented in the README).

| Gate item | Status |
| --- | --- |
| Reference app grows two organizations | ✅ `org-a`/`org-b` via the two pinned deployments; ensure-by-slug confirmed by probe |
| Org-A member against org-B's deployment: logged out on protected pages; sign-in rejected inline | ✅ Probe: 403 `MEMBERSHIP_REQUIRED`, **no session minted**; carve-out session resolves unauthenticated on every request. Inline error-code UX on the login pages; browser confirmation is walkthrough scenario 19. OAuth rejection surfaces via BetterAuth's OAuth error callback (manual, needs Google credentials) |
| Roles from `member.role`, incl. multi-role CSV; member row deleted mid-session loses access | ✅ Probe: `updateMemberRole(['admin','auditor','branch-manager'])` stores `"admin,auditor,branch-manager"`, splits back; mid-session member delete → unauthenticated on the next request |
| `_user.attributes` merges user + active member, member wins, shallow | ✅ Probe: `{region (user), branches (member wins), nested (replaced)}` |
| Open signup auto-joins with built-in `member`; unverified signup holds member row, no session | ✅ Probe (auto-join + role); unverified-holds-member-row is walkthrough scenario 18 (requires the email verification loop) |
| Invite → sign-up → accept; pending invitation admits pre-accept session; `contactId` stamped; expired invitation rejected | ✅ Probe: all four steps, including the stamp (`contactId=contact-123`) and expired-invitation 403 |
| Tenant: fresh signup mints own org lazily as owner; invited signup joins inviter's tenant, mints nothing | ✅ Probe: `org-<userId>` minted at first session, role `owner`, set active; invited signup minted nothing |
| Merge-on-signup via both bindings; invited user with stamped `contactId` skips merge | ⚠️ Implemented and build-validated in the reference app; runtime confirmation is walkthrough scenario 25 (needs the Mailpit verification loop and a Google OAuth signup). The skip-when-`contactId`-present guard is in both endpoints |
| `SetActiveOrganization` switches org; client sees new roles after `UpdateSession` | ⚠️ Implemented end to end (`organization.setActive` → `/api/user` re-sync); browser confirmation is walkthrough scenario 28 |
| Full suite passes: `pnpm test` | ✅ All packages green (build 1553, api 500, engine 349, actions-core 138, client 64, servers, …) |

## Verify-and-record (runtime probe, `better-auth@1.6.23`)

- **Multi-role CSV through `updateMemberRole` with empty-statement registered custom roles: works.** The array is flattened and stored as one CSV string; unregistered names are refused with `ROLE_NOT_FOUND`. The named limitation is confirmed live: a member holding only empty-statement catalog roles calling `inviteMember` gets 403 `YOU_ARE_NOT_ALLOWED_TO_INVITE_USERS_TO_THIS_ORGANIZATION` (the AC check needs `invitation: ["create"]`, which owner/admin carry). Last-owner protection also confirmed (`YOU_CANNOT_LEAVE_THE_ORGANIZATION_WITHOUT_AN_OWNER`).
- **`type: "json"` additionalField round-trip on MongoDB: works in code, but storage is a JSON *string*, not a sub-document.** The Mongo adapter does not set `supportsJSON`, which defaults to `false`, so the core field transform `JSON.stringify`s on write and parses on read (`@better-auth/core` `db/adapter/factory.mjs`). `mongodb/design.md` Decision 5 states a json field "maps to a native sub-document" — that does not hold at this version. Everything reading through BetterAuth (sessions, `resolveAuthentication`, admin steps) sees objects; a module reading `user-members`/`users` **natively** (which admin/design.md permits) sees strings and cannot aggregate on attribute contents. The #5900 typing bug is TypeScript-only — harmless here. **Recommendation:** feed back into mongodb/design.md; either accept string storage as the recorded convention or pursue `supportsJSON` upstream/in `MongoDBAuthAdapter`.

## Design learnings and deviations

1. **After-hooks flush at request end (1.6.23), so `user.create.after` auto-join runs *after* `session.create.before`.** An open signup minting an immediate session was rejected by its own policy hook. The pinned+open policy hook now also ensures membership at session-create; both joins are idempotent. Side effect worth recording: under pinned+open, a member removed by an admin who signs in again is auto-re-joined — consistent with open signup (they could re-sign-up anyway), but the designs may want to state it.
2. **The `createOrganization` endpoint cannot serve engine-tier calls at 1.6.23.** The engine's baseURL is dynamic (per-request host), and `resolveDynamicContext` only accepts the call's own `headers`/`request` as a source — a headerless system-action call throws, while forwarding the firing request's headers makes the endpoint demand a session that does not exist yet at `session.create`. Tenant orgs are therefore minted through the org plugin's own adapter layer (`getOrgAdapter`, exported from `better-auth/plugins`) — the same layer the endpoint drives. This skips the endpoint's slug-exists check (our slugs are `org-<userId>`, unique) and `organizationHooks.before/afterCreateOrganization` (nothing binds them in Lowdefy). Same reasoning for the startup ensure-by-slug seed, which additionally has no creator user at all (the endpoint always mints an owner member row).
3. **Nested `auth.api` calls from hooks must carry the firing request** (`getHookRequestHeaders`) or they cannot resolve the dynamic baseURL — `addMember` is called this way.
4. **`/api/user` added (servers).** The designs are silent on how the client sees member-resolved roles after `UpdateSession` (the base session carries neither roles nor the merged attributes), but the gate requires it. The servers now expose the resolved caller — exactly the `resolveAuthentication` output the first render embeds. Recommend folding this into the engine/config-schema designs.
5. **`SetActiveOrganization` under `pinned` — open design question.** `resolveAuthentication` reads `session.activeOrganizationId` per the design, and the org plugin's `setActive` endpoint is mounted for everyone. A user with memberships in several orgs can therefore switch their active org *within* a pinned deployment, resolving that deployment's pages against another org's roles. If the wall under `pinned` should be pinned to the configured org (or `setActive` blocked), that is a design decision to settle — not decided in code.
6. **`email.verified` merge write pathway** (already an open question in hooks/design.md): the reference app writes `user.contactId` over a Lowdefy connection with an explicit update. Flagged in config comments; settle with the auth module design.
7. **Magic-link rejection surfaces at link consumption**, not on the `signIn.magicLink()` call (the session is only created when the emailed link is verified). The design's "magic link inline" framing should be confirmed against the walkthrough; recorded in README scenario 19.
8. The build-side "coverage table" lives in `config-schema/design.md` (not in this repo); its `plugins` row already names `organizations`. No repo-side table exists to update.

## Verification

- Unit: `pnpm test` green across the monorepo (includes new tests for build validation/defaults/role catalog, the policy hook, auto-join, invitation stamp, ensure-by-slug, pending-invitation lookup, plugin assembly, `resolveAuthentication` wall/roles/attributes, and the `SetActiveOrganization` action).
- Runtime: 21/21 probe checks against `better-auth@1.6.23` + in-memory MongoDB through the real engine assembly (`getBetterAuth` → hooks → `resolveAuthentication`).
- All three reference apps build cleanly and boot under the dev server.


SamTolmay:

PR #2218, opened 2026-07-02
feat(auth): Phase 4 - API auth strategies (apiKey, jwt)

SamTolmay:
Phase 4 of the auth upgrade: API auth strategies per [api-strategies/design.md](https://github.com/lowdefy/lowdefy-design/blob/main/designs/auth-upgrade/api-strategies/design.md). External callers authenticate to Lowdefy API endpoints with a static API key (`X-API-Key`, configurable) or a Bearer JWT (HMAC or JWKS); strategies grant roles and the unchanged role check authorizes. Stacked on #2217 (phase 3) like the other phase PRs — phases 2–3 were already complete on this line, so the parallel-worktree option in the prompt no longer applied and the strategy loop integrates directly against the real phase-3 `resolveAuthentication` seam.

## What changed per package

- **`@lowdefy/errors`** — new `AuthenticationError` (name-keyed, `isLowdefyError`), exported and documented in the class hierarchy.
- **`@lowdefy/build`** — `auth.strategies` registered in the closed auth schema; new `buildAuthStrategies` validates ids unique, types known (`apiKey`, `jwt`), `apiKey` keys present, `jwt` `secret` XOR `jwksUri`, `algorithms` required and non-empty, and writes per-entry defaults (`roles: []`, `attributes: {}`, `headerName: X-API-Key`); `strategies.length > 0` now counts as an authentication mechanism, so a strategies-only block is validly configured with **no database and no login method**; strategy types flow through the plugin type system into a new `plugins/auth/strategies.js` artifact (counters, buildTypes, imports, writeAuthImports, typesMap plumbing); success fixture `95-auth-strategies-app` locks the strategies-only build.
- **`@lowdefy/plugin-better-auth`** — `apiKey` and `jwt` verifier factories exported as `auth/strategies` plugin types. `apiKey`: SHA-256-then-`timingSafeEqual` (constant-time, length-blind), per-key audit identity in the synthetic caller id, startup warning on short keys. `jwt`: `jose` (`^6.2.3`, latest) with HMAC or `createRemoteJWKSet` JWKS, required `algorithms` allowlist, `issuer`/`audience`, `claimMapping` with nested role paths and `attributes.*` paths. Package gains a real jest setup (22 tests, including a live local-JWKS verification).
- **`@lowdefy/api`** — `resolveStrategyCaller` tries strategies in config order when no session resolves (first match wins; session branch stays terminal — a walled-out member never falls through to strategies); caller shape `{ id, authMethod, strategyId, roles, attributes }` with roles = static ∪ claim-derived (deduped) and attributes = static ⊕ claim-mapped (claim wins); the authenticating strategy is logged (`auth_strategy_authenticated`); `createAuthStrategies`/`getAuthStrategies` build verifiers once at startup, resolving `_secret`; `getBetterAuthConfig` constructs BetterAuth **without** a database adapter for strategies-only apps (stateless mode — verified against `better-auth@1.6.23` source: no `database` option → memory adapter, so `getSession` runs guard-free) and skips pinned-org seeding with no database; `authorizeRequest`/`authorizeApiEndpoint` throw `AuthenticationError` for unauthenticated callers on protected routes, keeping the opaque "does not exist" for wrong roles.
- **`@lowdefy/server` / `@lowdefy/server-dev`** — `getStrategies` accessor beside `getAuth`; `apiContext` passes strategies into `resolveAuthentication`; `createErrorHandler` gains the early branch keyed strictly on `error.name === 'AuthenticationError'`: one warning line + 401, before `handleError` and `Sentry.captureException`; custom plugin types maps (incl. server-e2e's) carry the `auth.strategies` store. server-e2e's cookie caller is untouched (pre-resolved carve-out).
- **Fixtures** — `apps/auth-strategies`: strategies-only reference app (no database, no login method) with role-gated `Api` endpoints, a bundled mock IdP (`jwks-server.mjs`: JWKS + RS256 token minting incl. bad-token modes) and HMAC minter (`mint-jwt.mjs`), and a step-by-step README walkthrough. `apps/auth-reference`: strategies block, role-gated `partner-data` + public `whoami` endpoints, walkthrough scenarios 29–32 (session-wins, walled-out-session, claim-derived roles).

## Gate

- ✅ **Authenticates via `X-API-Key`, Bearer JWT HMAC, and JWKS** — verified live against the served `auth-strategies` app: `partner-data` 200 with the key (`caller: "apiKey:partner-access:acme"`), `service-data` 200 with an HMAC token (`caller: "service-1"`), and a JWKS RS256 token from the mock IdP resolving `strategyId: "external-idp"`.
- ✅ **A logged-in session wins over a presented API key** — unit-verified: with a resolved session the strategy verifier is never called, and a session rejected by the membership wall also never falls through (`resolveAuthentication.test.js`). The live two-credential walkthrough is reference-app scenario 30 (needs MongoDB, not available in this run's environment).
- ✅ **No credentials on a protected endpoint → 401, one warning line, no Sentry, no structured log** — verified live: body `{"name":"AuthenticationError","message":...}`, exactly one `⚠ Unauthenticated request: POST /api/endpoints/partner-data` line per request in the server log, zero stack traces or structured error entries.
- ✅ **Valid credentials, wrong roles → opaque "does not exist"** — verified live: partner key on the `api-user`-gated endpoint returns the existing opaque ConfigError, no 403.
- ✅ **Disallowed algorithm / expired / wrong `iss`/`aud` rejected** — verified live: HS384-against-[HS256], expired, wrong-issuer, wrong-audience, and (JWKS) rogue-key tokens all → 401.
- ✅ **Strategy attributes on `_user.attributes`; `id`/`authMethod`/`strategyId` per the design shapes** — verified live via the `caller` endpoint for apiKey (static bag) and JWKS jwt (claim-mapped `attributes.branches: ["west"]`, nested `realm_access.roles`).
- ✅ **Strategies-only fixture builds and serves with no database** — verified live end-to-end; also locked by build fixture 95 (auth.json with `_secret` preserved, no database, artifact emitted).
- ✅ **Build rejects invalid strategy configs** — `buildAuthStrategies.test.js`: both/neither of `secret`/`jwksUri`, missing/empty `algorithms`, missing/empty `keys`, key without `value`, duplicate ids, unknown types.
- ✅ **Full suite passes** — `pnpm test` green across the monorepo (build 1577, api 519, plugin-better-auth 22, servers, errors).

## Design learnings / deviations

- **`jose` lives in `@lowdefy/plugin-better-auth`, not `@lowdefy/api`.** The design's package list puts jose in `@lowdefy/api`, but the `plugins/auth/strategies.js` build artifact resolves types to plugin packages through the same import machinery as adapters/providers — so the verifier factories (the only jose consumers) live with the plugin type declarations, and `@lowdefy/api` hosts the strategy loop, startup construction, and caller assembly. Pinned `jose@^6.2.3` (latest; dedupes with better-auth's `^6.1.3` peer range).
- **Session callers carry no `authMethod: 'session'`.** The engine design's flow diagram annotates the session branch with `authMethod: 'session'`, but phase 3's session branch doesn't set it, and changing the session branch is explicitly out of scope here. Recommend settling in the designs whether the session caller should carry `authMethod`/parity fields — routines that "branch on `_user.authMethod`" currently see `undefined` for session users.
- **Key id fallback.** The design leaves the synthetic id unspecified for keys without an `id`; the key's array index stands in: `apiKey:<strategyId>:<index>`.
- **Short-key warnings** use the engine's existing 32-character bar (`auth.secret` precedent) and also cover a short `jwt` HMAC `secret`, which is equally a static key.
- **Observation for the designs:** static strategy `attributes`/config values keep their build `~k` markers, so raw endpoint JSON shows serializer-wrapped arrays (`{"~arr": [...]}`) to external callers — server-side values are real arrays and the Lowdefy client deserializes transparently, and object-level `~k` noise already exists on every endpoint response, but array wrapping changes the JSON shape an external (non-Lowdefy) consumer sees. Worth a design note on external response hygiene (predates this phase; strategies just make external callers first-class consumers).

### DIFF
--- a/apps/auth-reference/README.md
+++ b/apps/auth-reference/README.md
@@ -37,6 +37,9 @@ later phase grows this suite with a scenario.
   # Only for apps/auth-reference-tenant (its own database - tenants do not
   # mix with the pinned apps):
   LOWDEFY_SECRET_TENANT_DATABASE_URI=mongodb://localhost:27017/auth-reference-tenant
+  # Only needed for the phase-4 API strategy scenarios:
+  LOWDEFY_SECRET_PARTNER_KEY_ACME=partner-key-acme-0123456789abcdef
+  LOWDEFY_SECRET_JWT_SIGNING_SECRET=jwt-shared-secret-0123456789abcdef
   # The hook scenarios assert this exact value lands in the audit row:
   LOWDEFY_SECRET_HOOK_AUDIT_KEY=audit-secret-value
   # Only needed to exercise the OAuth scenario:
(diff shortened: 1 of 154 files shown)


SamTolmay:

PR #2236, opened 2026-07-03
Group mentions for TiptapMentionInput

JohannMoller:
## Summary

`TiptapMentionInput` renders every `@`-mention the same way today: a flat suggestion list and one link-coloured chip per option. This PR lets an app mark any mention option as belonging to an arbitrary, app-defined **group** (a role, a queue, a team, a segment). Group options render under menu-section headings, insert a distinctly coloured chip carrying a `data-mention-group` attribute, and show the group's current members in a hover popover. Which groups exist, their labels, colours, and members are all app data — the block hardcodes none. Also fixes a latent `getHref` bug so a nullish href renders `<span>` instead of `<a href="null">`.

Design: `designs/tiptap-group-mentions/design.md`.

## Changes

- **@lowdefy/blocks-tiptap**:
  - **Menu sections** — options with `tag.section` group under headings in the suggestion dropdown; a single flat keyboard index is kept so arrow/Enter navigation crosses section boundaries.
  - **Group chips** — an option with `tag.group` renders `class="tiptap-mention tiptap-mention-group"` + `data-mention-group="<group>"`, coloured inline from `tag.color` so the colour travels with saved HTML.
  - **Hover member popover** — a new `useGroupMembersPopover` hook shows `mentions.groupMembers` (a `{ group: [{ name, email }] }` map, which may load from a request) on chip hover in the live editor; it delegates listeners on the editor DOM and tears the popover down on edit/unmount.
  - **Configurable cap** — `mentions.limit` (default 5) caps suggestions, applied per section when sections are used, otherwise across the flat list.
  - **`getHref` fix** — the element is now chosen by the function's return value: non-nullish → `<a>`, nullish → `<span>` (no more `href="null"`).
  - Presentation stays on `tag` (`section`/`group`/`color`); `value` remains the app's opaque, emitted payload — the block never reads it for rendering.
- **docs** — the `TiptapMentionInput` reference page gains a group-mentions guide (`tag` fields, `mentions.limit`/`groupMembers`, the corrected `getHref`, emitted chip markup) with a worked example.

## Key Files

- `packages/plugins/blocks/blocks-tiptap/src/blocks/TiptapMentionInput/TiptapMentionInput.js` — `renderHTML` group class/attr/inline colour + href-value element choice; wires the popover hook.
- `.../TiptapMentionInput/suggestion.js` — per-section cap with flat fallback.
- `.../TiptapMentionInput/MentionList.js` — group by section, headings, flat keyboard index.
- `.../TiptapMentionInput/useGroupMembersPopover.js` — new live-editor hover popover.
- `.../TiptapMentionInput/itemSection.js` — shared `tag.section` helper.

## Breaking Changes

- `getHref` returning `null`/`undefined` now renders `<span>` instead of `<a href="null">`. This only affects configs whose `getHref` returned a nullish value (previously a bug); configs returning real URLs are unaffected. Note the test is nullish-only — returning `''` still yields `<a href="">`.

## Notes

- Options with no `tag.section`, `tag.group`, or `tag.color` render exactly as before (flat menu, plain `tiptap-mention` chip) — fully backward compatible aside from the href fix.
- `mentions.groupColors` was considered and dropped in favour of the single `tag.color` source: a group is almost always one chip, and a central map would be captured at editor mount and silently fail for colours loaded from a request. `tag.color` rides on the selected option, so it is always current. `groupMembers` stays a map precisely because it is read live at hover time.
- e2e specs were added (`TiptapMentionInput.e2e.spec.js`) covering section headings, the per-section cap, cross-section keyboard nav, group chip attributes/colour, the `<span>`-vs-`<a>` href behaviour, and the hover popover lifecycle. They were not executed in the authoring sandbox (Playwright browser download is blocked there) and are expected to run in CI.

### DIFF
--- a/.changeset/feat-tiptap-group-mentions.md
+++ b/.changeset/feat-tiptap-group-mentions.md
@@ -1,0 +1,14 @@
+---
+'@lowdefy/blocks-tiptap': minor
+---
+
+feat: Group mentions for TiptapMentionInput
+
+`TiptapMentionInput` can now treat mention options as app-defined **groups** (roles, teams, queues, segments) in addition to individual people. All group data — which groups exist, their labels, colours, and members — comes from your config; the block ships no groups of its own.
+
+- **Menu sections** — options with a `tag.section` are grouped under headings in the suggestion dropdown; options without one render flat, as before.
+- **Group chips** — an option with `tag.group` renders a distinct chip carrying `class="tiptap-mention tiptap-mention-group"` and `data-mention-group="<group>"`, coloured inline from `tag.color` so the colour travels with the saved HTML.
+- **Hover member popover** — `mentions.groupMembers` (a `{ '<group>': [{ name, email }] }` map, which may be loaded from a request) shows a group's current members when you hover its chip in the live editor.
+- **Configurable suggestion cap** — `mentions.limit` (default 5) caps results, applied per section when sections are used so a large section can't crowd out a small one.
+
+Also fixes a latent `getHref` bug: the element is now chosen by the function's return value, so a nullish return renders a plain `<span>` instead of `<a href="null">`. Options with no `tag.section`, `tag.group`, or `tag.color` are unaffected and render exactly as before.
--- a/packages/docs/blocks/input/TiptapMentionInput.yaml
+++ b/packages/docs/blocks/input/TiptapMentionInput.yaml
@@ -26,3 +26,8 @@ _ref:
     description: >
       Rich-text editor with @-mention support. Same feature set as TiptapInput,
       plus a mention dropdown populated from a static options list or a request.
+      Mention options can be marked as groups (roles, teams, queues) to render
+      under their own menu section, as distinctly coloured chips, and with a
+      hover popover listing the group's current members. See the guide below.
+    guide:
+      _ref: blocks/input/TiptapMentionInputGuide.md
--- a/packages/docs/blocks/input/TiptapMentionInputGuide.md
+++ b/packages/docs/blocks/input/TiptapMentionInputGuide.md
@@ -3,7 +3,7 @@
 Beyond the flat mention list, `TiptapMentionInput` can treat some mention options as **groups** — an app-defined kind such as a role, a team, a queue, or a saved segment. A group mention:
 
 - sits under its own heading in the suggestion menu (`tag.section`),
-- renders as a distinctly coloured chip (`tag.color` / `mentions.groupColors`), and
+- renders as a distinctly coloured chip (`tag.color`), and
 - shows the group's current members in a hover popover in the live editor (`mentions.groupMembers`).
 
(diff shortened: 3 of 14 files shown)


JohannMoller:
## Self-review

Posting my own review to save the next reviewer time and to flag where to focus. Verdict: **sound and internally consistent, no blocking bugs.** One pre-merge gate remains (CI e2e).

### Verified
- **Manually, on a locally-built app** (`blocks-tiptap` e2e app): sectioned menu with per-section cap, flat keyboard nav across the section boundary, group chip (`tiptap-mention-group` + `data-mention-group` + inline `tag.color`), the `getHref` fix (person → `<a href="/contacts?_id=…">`, group → plain `<span>`, no `href="null"`), and the hover member popover (header + members, torn down on edit).
- **Mechanically**: all example configs (gallery + tests + e2e) validated against `meta.js` (`additionalProperties: false`); block builds clean; no `groupColors` references remain in `packages/`.
- **tippy contract**: `useGroupMembersPopover` passes a single element to `tippy(...)`, which returns a single instance (6.3.7) — `.show()` is correct; contrast with `suggestion.js`'s `tippy('body', …)[0]` (array).

### Pre-merge gate
- **The e2e specs have not executed** — the authoring environment blocks the Playwright browser download. They're correct-by-construction and assert exactly the behaviours verified manually above, but please confirm the CI e2e job (`pnpm --filter @lowdefy/blocks-tiptap e2e`) is green before merge. The keyboard-nav test is intentionally coupled to the playground fixture (`limit: 3` + 4 People → 3rd ArrowDown lands on Finance); if that fixture changes, update the test.

### Design decisions worth a look
- **Presentation lives on `tag` (`section`/`group`/`color`), never on `value`.** `value` stays the app's opaque, emitted payload; the block never reads it for rendering. This is the core principle — worth confirming you're happy with it.
- **`groupColors` map was dropped** in favour of a single `tag.color` source. Rationale: a group is almost always one chip (so a central map rarely saved repetition), and a map read inside `renderHTML` is captured at editor mount and would silently fail for colours loaded from a request. `tag.color` rides on the selected option, so it's always current. `groupMembers` stays a map precisely because it's read *live* at hover time.
- **`getHref` element choice is by return value**, tested with `type.isNone` — so `''` still yields `<a href="">`. Callers must return `null`/`undefined` for a `<span>`. Called out in the schema description and docs.

### Minor / by-design (non-blocking)
- Dropdown shows `label`; chip shows `tag.title ?? label` — a differing `tag.title` makes menu and chip text diverge (intended; documented).
- Member with no `name` → empty name row (cosmetic). Popover dismisses on hover-out/typing, not on scroll/blur/Escape.

Backward compatible: options with no `tag.section`/`group`/`color` render exactly as before; the only behaviour change is the `href="null"` → `<span>` fix.

Gervwyk (approved):

PR #2238, opened 2026-07-03
Prevent Tiptap editor teardown crashes when navigating away mid-edit

JohannMoller:
## Summary

The `@lowdefy/blocks-tiptap` blocks contained three "act then immediately unmount/navigate" teardown races that all surface as `NotFoundError: Failed to execute 'removeChild' on 'Node'`. These were found and fixed in a downstream fork (`@mrmtech/plugin-tiptap`) and are ported back here since the source is identical.

The crash reproduces when the editor unmounts while something it owns is still torn down asynchronously — e.g. a user submits or starts editing content and immediately clicks away.

### Fixes

1. **Bubble/formatting menu** (`utils/PopoverMenu.js`) — replaced tiptap's `<BubbleMenu>` React wrapper with a manual `BubbleMenuPlugin` that renders the buttons into a self-owned container via a portal. The wrapper let its React-rendered menu element be `element.remove()`'d by the plugin on construction, so on unmount React tried to `removeChild` a node that had moved. Owning the container means React only ever removes the buttons from their real parent.

2. **`@`-mention popup** (`TiptapMentionInput/suggestion.js`) — `onExit` now destroys the React portal first and defers the tippy teardown to a microtask so it never runs synchronously inside React's unmount commit. Added optional-chaining guards to `onUpdate`/`onKeyDown`.

3. **Async image insert** (`TiptapInput/useTiptapState.js`, `TiptapMentionInput/useTiptapMentionState.js`) — bail if `editor.isDestroyed` after the async S3 upload resolves, instead of chaining commands onto a destroyed editor.

Also adds `@tiptap/extension-bubble-menu` as a direct dependency (previously only transitive via `@tiptap/react`), now that `PopoverMenu` imports it directly.

## Test plan

- Open a Tiptap editor, select text to show the formatting menu, then navigate away / unmount the block → no `removeChild` crash.
- Open the `@`-mention popup, then navigate away while it's open → no crash.
- Start an image upload, navigate away before it resolves → no crash against a destroyed editor.

## Notes

Verified downstream in production (the identical fork). Syntax-verified locally with swc; a full workspace build was not run in this worktree.

## PR Checklist
- [ ] Documentation updated if needed
- [ ] Self-review completed

### DIFF
--- a/.changeset/fix-tiptap-teardown-crashes.md
+++ b/.changeset/fix-tiptap-teardown-crashes.md
@@ -1,0 +1,12 @@
+---
+'@lowdefy/blocks-tiptap': patch
+---
+
+fix: Prevent rich-text editor crash when navigating away mid-edit.
+
+Fixed a "removeChild" crash in the Tiptap editor blocks that could occur when the
+editor was unmounted while the formatting menu or an `@`-mention popup was open, or
+while an image upload was still in progress — for example when a user submitted or
+started editing content and immediately navigated away. The editor now tears these
+down safely, so navigating away no longer throws or leaves the editor in a broken
+state.
--- a/packages/plugins/blocks/blocks-tiptap/package.json
+++ b/packages/plugins/blocks/blocks-tiptap/package.json
@@ -58,6 +58,7 @@
     "@lowdefy/blocks-antd": "5.3.0",
     "@lowdefy/helpers": "5.3.0",
     "@tiptap/core": "2.27.2",
+    "@tiptap/extension-bubble-menu": "2.27.2",
     "@tiptap/extension-file-handler": "2.27.2",
     "@tiptap/extension-highlight": "2.27.2",
     "@tiptap/extension-image": "2.27.2",
--- a/packages/plugins/blocks/blocks-tiptap/src/blocks/TiptapInput/useTiptapState.js
+++ b/packages/plugins/blocks/blocks-tiptap/src/blocks/TiptapInput/useTiptapState.js
@@ -54,6 +54,10 @@ function useTiptapState({ value, methods }) {
 
   const insertImage = async (editor, file, pos) => {
     const url = await s3FileUpload({ file, methods });
+    // The upload is async: bail if the editor was torn down meanwhile (e.g. the
+    // surrounding page navigated away before the upload resolved), otherwise the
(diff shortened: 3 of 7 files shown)


JohannMoller (commented):
## PR Review: #2238

**Linked Issues:** None found

Ports three well-understood "act-then-immediately-unmount" teardown races from the downstream `@mrmtech/plugin-tiptap` fork into upstream `@lowdefy/blocks-tiptap`. The source was byte-identical, the fixes are minimal and targeted, and DOM output is unchanged. No blockers.

### 🟡 Suggestions

**No regression test for the teardown paths**
`blocks-tiptap` has e2e infra (`tests/TiptapInput.e2e.spec.js`, `tests/TiptapMentionInput.e2e.spec.js`). These crashes only reproduce on unmount/navigation timing so they're genuinely awkward to cover, but a test that opens the formatting menu / `@`-popup and then unmounts the block would lock in the fix. Reasonable to defer, but worth a follow-up note.

### 🟢 Nits

- **`insertImage` is now duplicated verbatim** across `TiptapInput/useTiptapState.js:55` and `TiptapMentionInput/useTiptapMentionState.js:62` — including the new guard. Pre-existing duplication and out of scope here, but a candidate for extraction into `utils/`.
- **Changeset fans out to ~66 packages** (per changeset-bot). Expected given Lowdefy's locked `5.3.0` versioning — flagging only so it isn't a surprise at release.

### What's verified

- ✅ Syntax: all 4 edited files transpile cleanly with the repo's `swc`.
- ✅ Call sites (`TiptapInput.js:171`, `TiptapMentionInput.js:229`) unchanged — still `{!disabled && <PopoverMenu editor={editor} …/>}`.
- ✅ DOM/CSS contract preserved: portal renders into a `.tiptap-popover` container with `.tiptap-icon` children, identical to the old `<BubbleMenu>` wrapper.
- ✅ `useEffect` deps `[editor, container, hasTools]` are complete (no exhaustive-deps violation); `container` is stable via `useState`.
- ✅ `@tiptap/extension-bubble-menu@2.27.2` pinned to match the other `@tiptap/*` deps.
- ⚠️ Full workspace build not run in the worktree (no `node_modules`); logic proven in downstream production.

---

*Generated by PR Review Command*

Gervwyk (approved):

PR #2246, opened 2026-07-07
feat(build): module-level agents — scoped ids, LLM-safe tool names, _module.agentId

StephanieJKS:
## Summary

Modules can now ship agents. `module.lowdefy.yaml` gains an `agents:` section, processed like `connections`/`api` — agent ids scope to `{entryId}/{agentId}` and merge into `components.agents` before `buildAgents` runs. This is the framework prerequisite for the AI chat reporting module (design: `lowdefy-design/designs/ai-chat-reporting`); without it a module whose entire purpose is an agent cannot ship its agent.

## Changes

- **`buildModules`**: iterate `agents` alongside `connections`/`api` — validate module secret references, scope ids per entry, merge into `components.agents`. Multi-instance installs get distinct agents per entry.
- **`deferredRegions`**: `agents` joins the content-section regexes, so the header walk leaves the section raw and the full manifest walk resolves it with the module entry in scope (`_module.*` operators work anywhere in the agent config).
- **`registerModules`**: null-filter `agents` entries like the other sections.
- **Walker**: new `_module.agentId` operator (string and `{ id, module }` forms), mirroring `resolveModuleEndpointId`. Needed by `AgentChat.agentId` in module pages and `CallAgent` steps in module endpoints — no existing `_module.*` operator produces a scoped agent id.
- **Tool names** (`buildAgents` + `buildAgentTools`): endpoint tools and sub-agent references gain a model-facing `name`. Providers require `^[a-zA-Z0-9_-]{1,64}$` and scoped ids contain `/`, so the default name replaces `/` with `__`; an explicit `name` overrides (better LLM ergonomics: `query_data` instead of `reporting__query-data`). Validated for pattern, reserved platform names, and per-agent uniqueness across tools and sub-agents. `buildAgentTools` keys tools by name; execution still targets the `endpointId`.
- **Schema**: `name` allowed on tool and sub-agent objects.
- **AgentChat transport**: URL-encode the agentId path segment — the agent route takes the last segment as the agentId (`getPathSegments` decodes per segment), and scoped ids contain `/`.

## Tests

- New fixture `95-module-agents`: a module ships its agent — the snapshot pins scoped agent/connection/hook ids, an explicit tool name, the `helper__count-users` default, and `_module.agentId` resolution in both a page property and a `CallAgent` step.
- New unit tests for name defaulting, explicit names, invalid/reserved/duplicate names, and sub-agent name collisions; runtime tests for name-keyed tools.
- Fixture 94's snapshot picks up the defaulted tool name (`name: lookup-signups` — identical to the endpoint id for unscoped ids, so no behavior change for existing apps).
- Full suites green: 1,681 build tests, 166 ai-utils tests.

## Notes

- Verified end-to-end by building a test app that installs the reporting module from `lowdefy/modules-mongodb` (companion PR there): agent ids, tool endpoints, hooks, `CallAgent` agentId, Dynamic `endpointId` and client `_api` keys all resolve to entry-scoped ids.
- Targets `vite-hono` — the change builds on the module system and agent system on that line.

https://claude.ai/code/session_017jGT2SkTpc2JSkKAeT3vYJ

### DIFF
--- a/packages/build/src/build/buildAgents.js
+++ b/packages/build/src/build/buildAgents.js
@@ -24,6 +24,9 @@ import { RESERVED_PLATFORM_TOOL_NAMES } from '@lowdefy/ai-utils';
 import countOperators from '../utils/countOperators.js';
 import createCheckDuplicateId from '../utils/createCheckDuplicateId.js';
 
+// Provider tool-name rule (Anthropic and OpenAI both).
+const TOOL_NAME_REGEX = /^[a-zA-Z0-9_-]{1,64}$/;
+
 function detectCycles(agents) {
   const graph = {};
   for (const agent of agents) {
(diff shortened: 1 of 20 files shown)


machielvdw:
0.0.0-experimental-20260707145139

PR #2259, opened 2026-07-13
fix(blocks-antd-x): Don't replay agent events for restored conversation history

Gervwyk:
## Summary
- Loading a saved conversation into `AgentChat` (via `properties.messages` or the `setMessages` method) replayed `onToolCall`, `onToolResult`, `onUserMessage`, and `onTitleGenerated` for every historical message — re-running side effects (state updates, API calls) that had already happened the first time. `useAgentEvents` now recognizes externally-loaded messages via `collectExternalEventIds` and suppresses those four events for their ids; only genuinely new activity fires them. Dedup bookkeeping (`firedToolCallIds` etc.) is unchanged, so this is purely about which ids get to *trigger* an event, not which get *tracked*.
- Separately: `onToolCall` fired as soon as a tool part appeared, even while its `input` was still streaming in (`part.state === 'input-streaming'`) — so consumers wiring `onToolCall` to apply a tool's structured output could receive a truncated object. `onToolCall` now waits for the input to finish streaming.
- Found and fixed while building an app that restores per-block AI-assistant conversations and applies the assistant's tool output to a canvas — restoring a thread was silently re-applying old tool calls (and once, replaying a truncated one) on top of fresh, unrelated edits.

## Test plan
- No existing automated coverage for this block; `packages/plugins/blocks/blocks-antd-x` has no `*.test.js` files today.
- `pnpm --filter @lowdefy/blocks-antd-x build` — clean (SWC compiles cleanly against `main`).
- Node smoke test of the compiled `collectExternalEventIds` export against a representative message set (user message, tool call with `output-available`, `data-chat-title` part) — ids collected correctly for all four suppressed event types.
- Manually exercised inside a consuming app: restoring a conversation with a completed `set-block-config` tool call no longer re-fires `onToolCall`; new tool calls in the same session still fire normally.


machielvdw:
This seems to include `develop <- main`  commits as well.

machielvdw (approved):
The changes in 9561b397a64b388b52b27df39fb519c81548c89e look good

PR #2277, opened 2026-07-15
feat(connections): Mail send requests return per-message results with post-filter to

machielvdw:
## What

`SendGridMailSend` and `SMTPMailSend` now return a `results` array with one entry per message sent, each including the post-filter `to` — the address mail was actually delivered to after the connection `filter` (`replaceAddress` / `allowlist` / `regex`) is applied.

- **SendGrid**: `SendGridMailSend` previously discarded `send()` results and returned only `{ response: 'Mail sent successfully' }`. It now collects per-message results like `SMTPMailSend` already did, so the SendGrid `messageId` finally reaches routines.
- **Both plugins**: `send()` returns `to: filtered.to` on success, and `{ messageId: null, to: null, filtered: true }` when the filter drops the message entirely.
- `applyMailFilter` is untouched in both plugins — the final `to` is read off its existing return value.

## Why

Notification routines persist a record per email (e.g. the modules-mongodb notifications module). With a `replaceAddress` dev-redirect active, the record's intended recipient and the actual delivered-to address silently diverge, and nothing in the data shows the redirect fired. Surfacing the post-filter `to` lets routines record both — intended (`email`) and actual (`email_result.to`).

## Tests

- Updated SendGrid `SendGridMailSend.test.js` result assertions (now `results` array), added filtered-drop and `replaceAddress` redirect cases.
- Updated `send.test.js` in both plugins for the new `to` field, added redirect cases.
- Updated `SMTPMailSend.test.js` mock + assertions for the `to` passthrough.
- `pnpm --filter=@lowdefy/connection-sendgrid test`: 55 passed. `--filter=@lowdefy/connection-smtp test`: 47 passed. Both at 100% statement coverage; both packages build.

Changeset included (minor, `@lowdefy/connection-sendgrid` + `@lowdefy/connection-smtp`).

### DIFF
--- a/.changeset/feat-mail-send-results.md
+++ b/.changeset/feat-mail-send-results.md
@@ -1,0 +1,13 @@
+---
+'@lowdefy/connection-sendgrid': minor
+'@lowdefy/connection-smtp': minor
+---
+
+feat: Mail send requests return per-message send results
+
+`SendGridMailSend` and `SMTPMailSend` now return a `results` array with one entry per message sent, so routines can record delivery outcomes:
+
+- Each result includes the post-filter `to` — the address mail was actually delivered to after the connection `filter` (`replaceAddress`, `allowlist`, `regex`) is applied. When a filter redirects mail to a test inbox, routines can now persist both the intended recipient and where the message really went.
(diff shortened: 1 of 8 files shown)


machielvdw:
0.0.0-experimental-20260715105919

PR #2280, opened 2026-07-17
Tenant wall — multi-tenant data scoping for MongoDB connections

machielvdw:
## Summary

Implements the **tenant wall** from `designs/auth-upgrade/_completed/mongodb-data-scoping/design.md` + amendments 1–3 (lowdefy-design, `main`): under `auth.organizations.policy: tenant`, every connection whose type implements the scoping contract is walled to the caller's active organization — enforced mechanically on reads **and** writes by the MongoDB connection, computed per request by the engine, and fail-closed for callers with no organization (system context, strategies) unless the request declares the explicit `tenant: none` opt-out.

**The policy sets the scoping default; a connection declares only its exception** (amendment-3). A scoped connection carries no tenancy key at all; `tenant: shared` is the one connection-level path out of the wall, for data deliberately shared across organizations, and `tenant: { field }` scopes on a non-default field. `tenant: true` is removed — it would restate the default — and is a build error naming its replacement. The rule applies uniformly to module-owned and app-owned connections.

The wall is **policy-conditional** (amendment-2): it engages only under `policy: tenant`. Under `pinned` (the default) no connection is scoped and none can be — no read filter, write stamp, selector merge, change-stream scoping, authored-clause audit, fail-closed error, or authored-field rejection — so a single-org deployment adopts scoping-capable connections with no backfill or index work, and `tenant: shared` is inert. The flip to `tenant` is the deliberately breaking step, and a serving preflight makes it loud rather than blank.

This is the framework prerequisite for the org-aware-modules design in modules-mongodb, and covers that design's blocking upstream ask (Atlas `$search` on walled collections) through the `tenant: authored` audit contract.

## Changes

**@lowdefy/api**

- `context.user.organizationId` exposed (readable as `_user: organizationId`, both org policies).
- `resolveTenant` produces the `{ field, value }` verdict: `null` under `policy: pinned`, on a `tenant: shared` connection, on a non-scopable connection type, or under a request-level `tenant: none`; `authored: true` under `tenant: authored` (the verdict still resolves — org-less callers are still rejected); fail-closed otherwise. Threaded through `callRequestResolver`, routine steps, webhook verify, and the websocket seam — where the verdict also joins the channel identity hash so two orgs never share a change-stream source.
- `resolveTenantPreflight`: the flip preflight. Under `policy: tenant` the server refuses to serve while any walled collection holds rows missing the tenant field (or carrying `null`), naming every offending collection in one aggregated `ConfigError`. Run lazily once per process and awaited in the api-context middleware; a refusal memoizes until restart, probe connectivity failures retry.

**@lowdefy/connection-mongodb**

- Selector `$and`-merge on every find/update/delete, insert/replace stamping, explicit upsert stamping (`$setOnInsert` / appended pipeline `$set`), bulkWrite per-op dispatch, and change-stream `fullDocument` matching with forced `updateLookup` (events that cannot prove they match — deletes — are not delivered).
- Aggregation: the wall's one pipeline move is prepending the tenant `$match` at pipeline entries — root, `$lookup` (both forms), `$unionWith`, recursing through `$facet`. It never rewrites the inside of a stage. `$out`/`$merge` and `$collStats`/`$indexStats` are rejected on tenant connections.
- Stages the prepend cannot scope — a pipeline-leading `$search`/`$searchMeta`/`$vectorSearch`/`$geoNear`, and `$graphLookup` at any position — are **refused** unless the request declares `tenant: authored`, in which case the request authors the tenant clause itself and the wall **audits** it before the driver call: clause present, tenant field, value strictly equal to the resolved verdict, in an AND position only (`$or`/`$nor` do not qualify). Any miss refuses to run. The rest of an authored request's pipeline is still walled mechanically.
- Loud rejection of authored tenant-field usage in write and filter positions (reads of the field stay allowed); the audited stage of a `tenant: authored` request is the one place the field is instead required.
- `tenantPreflight` implemented as a connection-type capability for the flip probe.

**@lowdefy/build**

- The connection position accepts only the exception: `tenant: shared | { field }`; `tenant: true` is a build error naming its replacement, and `none`/`authored` on a connection errors naming the request position (the two positions keep disjoint value sets in both directions). Any declaration on a type without the scoping contract stays a build error under both policies, so an unenforceable declaration fails on the current deployment rather than at flip time.
- Connection types declare their tenancy capability via the `connectionMetas` capability map (the connection analog of `blockMetas`): `{ tenant: true }` implements the scoping contract (v1: `MongoDBCollection`), `{ tenant: false }` is non-scopable. Because silence now means scoped, a connection whose type declares neither is a build error under `policy: tenant` — a data-bearing custom type can never be silently unscoped. All 17 first-party connection types now declare one or the other.
- `tenant: none | authored` sentinels on requests/steps (websockets accept `none` only), the new `tenantConnections.json` artifact enumerating the walled set (every scoping-capable connection not declaring `shared`), and `validateTenantPipelineEntry` — a best-effort literal-pipeline entry-stage check, active only under `policy: tenant` since runtime remains the enforcement gate for operator-composed pipelines.
- `validateTenantRemaps`: a connection remap swaps the module's whole connection definition for the app's, wall position included. Under `policy: tenant`, remapping a scoped module connection onto a target that is **not** scoped — one declaring `tenant: shared`, or one whose type lacks the scoping contract — is a `ConfigError` naming both connections; otherwise the module's plain reads run unfiltered and its writes unstamped while its `tenant: authored` requests keep their clause, so list pages look scoped while everything underneath leaks. A target that declares nothing is scoped like any other connection, so plain remaps stay legal.
- `_build.authConfig` projects `organizations.policy` so config can branch on the policy at build time, and a new `deferAuthConfig` walk flag lets the operator resolve in module entry vars and cross-module components consumed through them: those walks run before the auth-config projection exists, and their output is always re-walked at the consuming site, so the fold can wait instead of erroring. The `auth:` self-reference and app-metadata positions still throw, with the boundary message reworded to name only those. Error fixture `B7` becomes success fixture `99`; new fixture `98` covers the cross-module component chain through entry vars.

**@lowdefy/server / server-dev / server-e2e**

- `connectionMetas` store in the custom-plugin types-map skeletons; api-context middleware awaits the preflight; the server-e2e and server-dev mock/headless contexts (which run no auth engine) derive the organizations-policy binding from the built auth config, so the wall is not silently disengaged for tenant apps there.

## Key Files

- `packages/api/src/routes/request/resolveTenant.js` — verdict production, policy gate, fail-closed rejection
- `packages/api/src/routes/connections/resolveTenantPreflight.js` — flip preflight refusal
- `packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/tenant/injectTenantIntoPipeline.js` — prepend-only injector, refuse/audit dispatch
- `packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/tenant/auditAuthoredClause.js` — runtime audit of an authored tenant clause
- `packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/tenant/applyTenantToUpdate.js` — update guarding + explicit upsert stamping
- `packages/build/src/build/buildConnections.js` — connection-type capability gate, `tenantConnectionIds`
- `packages/build/src/build/buildModules.js` — `validateTenantRemaps`
- `packages/plugins/connections/connection-mongodb/src/connections/MongoDBCollection/tenantIsolation.test.js` — two-org isolation proof per resolver against real MongoDB

## Notes

- **Tenant only** — attribute scopes (`scope:`) are deferred to a later PR per the design's milestone open question; the verdict seam takes `scope` as a sibling field when it lands.
- **Enabling the wall.** Adopting scoping-capable connections under `pinned` needs no migration and no declaration. The flip to `policy: tenant` does: the tenant field must be backfilled on existing documents (the preflight refuses to serve until it is), and every system-context caller (webhook verifiers, hook routines, scheduled endpoints) and strategy caller of a walled connection starts failing until it either runs with an organization or declares `tenant: none` at the point of use. `tenant: none` is inert under `pinned`, so authoring it ahead of the flip is safe.
- Consumer requirements for authored Atlas clauses: the tenant field `token`-mapped in the Atlas Search index (and in `storedSource` where `returnStoredSource` is used), and indexed as the `filter` type for Atlas Vector Search.
- `packages/docs` is deliberately untouched — the design's task 09 (documentation) covers it.
- Rides a latent-bug fix: `MongoDBChangeStream` called `.watch()` on the `getCollection` wrapper instead of the collection (masked by its test mock).
- One changeset (`feat-tenant-wall.md`) describes the final behaviour — the wall is unshipped, so amendment-1 and the policy gate are folded in place rather than added as follow-ups.
- **Design coverage**: the policy gate, flip preflight and remap guard are designed in amendment-2, the inverted default and the type-capability contract in amendment-3 (both in lowdefy-design `main`; amendment-3's open question — auto-exempting the auth adapter's collections — is settled **no**: no sound build-time key exists, and a carve-out would be a second, invisible path out of the wall). The `_build.authConfig` deferral remains implementation-decided with no design section (it also reverses `_completed/authconfig-module-projection` task 03, which named module entry vars an explicit non-goal).
- Tests green on `@lowdefy/api` (963), `@lowdefy/build` (2065), `@lowdefy/connection-mongodb` (457), and the 16 plugin packages that gained non-scopable capability declarations. The `B6-auth-config-unknown-path` failure noted earlier in this PR was a stale-dist artifact and no longer occurs.


machielvdw:
0.0.0-experimental-20260717134857

machielvdw:
0.0.0-experimental-20260720072521

machielvdw:
0.0.0-experimental-20260730112404

machielvdw:
0.0.0-experimental-20260803075850

machielvdw:
0.0.0-experimental-20260803084426

machielvdw:
0.0.0-experimental-20260804101357

packages/servers/server/lowdefy/createCustomPluginTypesMap.mjs
```diff
@@ -46,6 +46,7 @@ async function createCustomPluginTypesMap({ directories }) {
       strategies: {},
     },
     blockMetas: {},
+    connectionMetas: {},
```
machielvdw, line 49: What is this?

packages/api/src/context/normalizeInjectedCaller.js
```diff
@@ -22,12 +22,23 @@
...
 // a real user-row/member read, so they are left exactly as the caller carries
 // them - absent stays absent, and a bare _user.profile read is null-safe.
+// organizationId (the tenant wall's org value) mirrors resolveAuthentication,
+// where it always equals activeOrganizationId - an injected caller declaring
+// either key gets both, so mock/e2e callers pass walled connections without
+// duplicating the org under two names. A caller with neither stays org-less
+// and fails closed on walled connections, same as a strategy caller.
 function normalizeInjectedCaller(user) {
-  return {
+  const normalized = {
```
machielvdw, line 31: ???

packages/api/src/context/resolveAuthentication.js
```diff
@@ -97,6 +97,12 @@ async function resolveAuthentication(context, { auth, headers, strategies }) {
       ...(member.attributes ?? {}),
     },
     activeOrganizationId,
+    // organizationId is the caller's active org in its serialized string form -
+    // the value the tenant wall stamps onto and filters walled collections with,
+    // and the one operators read as _user: organizationId. It resolves under
+    // both organizations policies (under pinned it always equals the pinned
+    // org, since set-active-organization is disabled there).
+    organizationId: activeOrganizationId,
```
machielvdw, line 149: why the two filelds? there is one above on line 99? Whats the difference?

packages/build/src/build/buildRefs/deferredRegistry.js
```diff
@@ -188,6 +188,11 @@ async function resolveRecordBody(ctx, id, record) {
     env: ctx.env,
     lowdefyApp: ctx.lowdefyApp,
     dynamicIdentifiers: ctx.dynamicIdentifiers,
+    // A record demanded before the auth-config projection exists (the entry
+    // sweep) memoizes its value, and that value is re-walked wherever it is
+    // consumed — so _build.authConfig folds defer instead of erroring. Once
+    // the projection exists the guard never fires and folds run here.
+    deferAuthConfig: true,
```
machielvdw, line 195: What is this?

machielvdw:
# Fable review

Three parallel review passes (MongoDB enforcement, API plumbing, build side), every finding below re-verified in the code. The core design held up under adversarial checking — the `$and` filter merge, upsert stamping, pipeline injection, bulkWrite whitelist, and fail-closed plumbing are sound — but there is one real hole and a few things to fix before merge.

> **Update:** all four must-fix items are fixed and pushed — 2585034f5 (1), a0860d6cf (2), 724d4b734 (3), f66e55a83 (4) — plus f16ad7a14 for the prettier failures. Each fix carries new tests; full suites green in `@lowdefy/build` (2069), `@lowdefy/api` (968), `@lowdefy/connection-mongodb` (463). The "should fix or decide" and remaining smaller items are still open.

## Must fix

**1. ✅ Fixed in 2585034f5 — A dotted tenant field breaks the wall — including a cross-tenant write escape.** `tenant: { field: 'meta.organizationId' }` passes build (`buildConnections.js:114` accepts any string), but every tenant helper treats the field as an opaque key:
- The authored-field guard (`assertTenantFieldNotAuthored.js:43`) matches keys against the whole string, so `{ $set: { meta: { organizationId: 'org_b' } } }` sails through — the walled filter matches the caller's own row, then the `$set` moves it into org B. `$unset: { meta: '' }` unstamps a row the same way, and `$rename` onto/away from the parent path also passes (`applyTenantToUpdate.js:57-67`).
- The insert stamp (`stampTenantOnDoc.js:32`, `{ ...doc, [field]: value }`) writes a **literal** `"meta.organizationId"` key, which the path-based read filter never matches — inserted docs become invisible to every walled read, while upserts and pipeline updates write the real nested path.

Simplest fix: reject a `field` containing `.` in the schema and `buildConnections`; no test uses one. Same check should reject `''` — `type.isString('')` passes today and the schema's `minLength: 1` is warning-only.

**2. ✅ Fixed in a0860d6cf — changeLog documents are written unstamped.** All nine log inserts (e.g. `MongoDBInsertOne.js:41`, `MongoDBUpdateOne.js:66`) write the log doc outside the stamp path, so a walled connection's audit trail carries one tenant's `before`/`after`/`payload` with no tenant marker. Worse: if the changelog collection is itself reachable through a walled connection, the preflight probe (`tenantPreflight.js:33`) finds the unstamped rows and `resolveTenantPreflight` memoizes a permanent refusal — and a backfill can't fix it because the next write recreates unstamped rows. The log doc should be stamped with the same verdict as the triggering operation.

**3. ✅ Fixed in 724d4b734 — Webhook verify: the tenant failure escapes its documented fail-closed path.** `runWebhookVerify.js:61` calls `resolveTenant` *outside* the `try` at line 71, but the comment at 54–56 promises "caught below, verdict false". The `AuthenticationError` propagates to the prod error handler, which returns the message — internal connection id plus tenant-wall remediation advice — in a 401 body to the unauthenticated webhook sender, and the `webhook_verify_failed` warn event never fires. Move the call inside the try (or catch it into a false verdict).

**4. ✅ Fixed in f66e55a83 — Tenant preflight runs with the first caller's identity, contradicting its own contract.** The file's comment says connection properties evaluate "caller-less … caller operators resolve to nothing", but `resolveTenantPreflight.js:68` calls `createEvaluateOperators(context)`, which captures `context.user` (`createEvaluateOperators.js:20`) — and in both servers the preflight runs after `resolveAuthentication`. A connection whose properties read `_user` gets probed against whichever caller hits the cold process first, and the pass/refuse verdict is memoized for the process lifetime. Evaluate with a nulled user.

## Should fix or explicitly decide

**5. Websocket tenant verdicts are frozen at upgrade time.** The upgrade handler captures the context once and every later `subscribe` frame resolves its verdict from that snapshot (`prepareChannel.js:54`), so an org switch or membership removal never affects an open socket — it can keep subscribing and receiving walled change events as the old org indefinitely. This breaks `resolveAuthentication`'s stated "member row read live on every request" invariant. Auth-at-connect is a defensible websocket semantic, but then it should be a documented decision plus a teardown/re-check story (e.g. close sockets on set-active-organization).

**6. The consecutive-id counter reads across the wall.** `getConsecutiveIdIndex.js` aggregates over the whole collection with no tenant clause, so tenants share one `T000042` sequence and can infer other orgs' record counts. Possibly intentional for `_id` uniqueness — but it's an unwalled read on a walled connection and deserves a stated decision (or a per-tenant prefix).

**7. Dev/prod divergence: the JIT dev server never runs the pipeline-entry check for page requests.** `context.tenantConnectionIds` is only created by `buildConnections` (`buildConnections.js:155`); nothing restores it in the JIT page-build path, so `validateTenantPipelineEntry.js:35` silently skips. A dev adds a literal `$search` pipeline, the dev server serves it happily, and CI's `lowdefy build` then fails — the check is absent exactly where authors iterate. (Runtime still audits, so it's DX, not a hole.) Persist/restore the set like `connectionIds`.

**8. The `deferAuthConfig` premise is false for module-entry connection remaps.** `buildModuleDefs.js:89-93` defers folds for both entry sections on the claim that output is re-walked through `_module.var` — true for `consumerVars`, not for `connections`: remap values are consumed directly (`readRemapValue` at `walker.js:445` splices deferred *records* but never re-walks folds). A policy-branching remap like `connections: { db: { _build.if: … _build.authConfig … } }` leaves a raw operator object as the target id: `validateTenantRemaps` (`buildModules.js:171`) silently no-ops on it and the build eventually dies with a baffling "connectionId is not a string. Received {…}". Re-walk `moduleEntry.connections` post-projection, or error loudly on a leftover fold.

**9. Design note worth an explicit statement:** the default types map declares `tenant: false` ("non-scopable") for Knex, Elasticsearch, Redis, and Google Sheets — queryable datastores lumped in with SMTP. Under `policy: tenant` an app serves cross-tenant SQL/ES data with zero build signal. MongoDB-only-v1 is fine, but the changeset/docs should say that plainly.

## Smaller

- `buildConnections.js:136` errors only on `undefined` capability meta, while every walled-set predicate requires `=== true` — a plugin exporting `tenant: 'true'` (string) passes build, is excluded from `tenantConnections.json`, then fails per-request at runtime. Treat non-boolean as undeclared.
- `validateTenantRemaps` reads `components.auth?.organizations?.policy` at `buildModules` (index.js:126) before runtime-operator folding (line 135), so a statically-foldable `policy` value makes the remap guard and the wall disagree.
- `writeConnections.js:20` early-returns on a zero-connection app before writing `tenantConnections.json`, contradicting its own "written under both policies" comment — the server then logs the misleading "rebuild with a matching version" preflight warning forever.
- Change streams silently force `fullDocument: 'updateLookup'` and drop delete/invalidate events (`applyTenantToChangeStream.js:47`) — quietly-incomplete feed instead of a refusal for a subscriber authored for deletes; also no integration test passes `tenant` to `MongoDBChangeStream`.
- `tenant: none` on non-request routine steps (CallApi etc.) is silently accepted and ignored — `validateStep.js` returns before the sentinel check.
- Request `options` pass through to the injected tenant equality, so caller-chosen `collation` (e.g. `strength: 1`, `numericOrdering`) governs the wall's comparison. Only exploitable if two org ids collide under the collation — hardening, not a live bypass.
- ✅ Fixed in f16ad7a14 — Prettier fails on 4 new files in `connection-mongodb/…/tenant/`: `applyTenantToBulkOperations.js`, `injectTenantIntoPipeline.js`, and two test files.
- `auditAuthoredClause.js` — the acceptance gate for `tenant: authored` — has no direct unit test; `normalizeInjectedCaller.js:37` uses `!= null` where the repo standard is `type.isNone()`.

## Verified clean

No unwalled executor path exists (all three request executors resolve tenant first; agents/MCP go through walled routes); no client-controlled value reaches the verdict; the `$and` filter merge cannot be disjuncted away; upsert stamping covers object, pipeline, and query-extraction paths including `$replaceRoot`/`$arrayToObject` tricks; the pipeline injector recurses every sub-pipeline form and refuses `$out`/`$merge`/`$collStats`/`$indexStats`; bulkWrite's schema (`maxProperties: 1`) plus kind whitelist closes multi-key ops; websocket channel keys hash the tenant so orgs never share a running source; the deferral core is coherent for the `consumerVars` path with correct B5/B6/B7 error behavior.

packages/servers/server/lowdefy/createCustomPluginTypesMap.mjs
machielvdw: `connectionMetas` is the new per-connection-**type** capability store in the types map — metadata a plugin exports from its `types.js`, parallel to `blockMetas` on the line above. Concretely, connection-mongodb now exports `connectionMetas: { MongoDBCollection: { tenant: true } }` ("this type implements the tenant scoping contract").

Build consumes it in three places: `buildConnections` validates `tenant:` declarations against it (and under `policy: tenant` errors when a data-bearing type declares neither `true` nor `false`), and `writeConnections`/`buildModules` use it to enumerate the walled set into the `tenantConnections.json` artifact for the server preflight.

This line just adds the empty bucket to the custom-plugin skeleton so custom plugins' metas get collected too (same one-line change in the server-dev and server-e2e copies). Runtime enforcement doesn't read this map — it reads the parallel `meta: { tenant: true }` on the connection object itself (`MongoDBCollection.js`).

packages/api/src/context/normalizeInjectedCaller.js
machielvdw: Three paths inject a user **without** going through `resolveAuthentication` (the real session → member-row resolution): the dev server's mock user, the dev headless-renderer cookie user, and the e2e test cookie user. Each substitutes for the whole resolve step, so each must produce the same user shape — otherwise `_user` reads and the tenant wall behave differently in dev/e2e than in prod.

This function is the single place that shape is floored: `roles` → `[]`, `attributes` → `{}` (so a whole-bag `_user: attributes` read gives `{}` like a real caller), and — the tenant-wall part — org mirroring: an injected caller declaring either `organizationId` **or** `activeOrganizationId` gets both, so a mock user with one org key passes walled connections in dev; declaring neither stays org-less and fails closed exactly like prod.

Call sites: `createLowdefyContext.js:103,111` (server-dev mock + headless) and `server-e2e/src/middleware/apiContext.js:80`.

packages/api/src/context/resolveAuthentication.js
machielvdw: They're an alias — always equal for a real session — but the two names come from different eras:

- **`activeOrganizationId` (line 99) predates this PR**: BetterAuth's session field name, already documented public `_user` surface in the auth-upgrade migration doc (`packages/docs/migration/auth-upgrade.yaml:212`).
- **`organizationId` (line 105) is new here**: the tenant wall's canonical name for "the caller's org". `resolveTenant.js:91` reads `context.user.organizationId`, and the name deliberately matches the default document field the wall stamps and filters — so an authored tenant clause reads `_user: organizationId` and lands in the doc field of the same name. "Active" would also be a misnomer under the pinned policy, where org switching is disabled.

Worth noting: nothing in the repo reads `context.user.activeOrganizationId` anymore — it survives only because the migration doc promises it. Since auth-upgrade is also unreleased, we could still consolidate to `organizationId` only and update that migration table. Happy to make that change if you prefer one name.

packages/build/src/build/buildRefs/deferredRegistry.js
machielvdw: `_build.authConfig` folds values out of the auth-config projection, which only exists once `resolveAuthConfigProjection` (buildAuth) has computed it. But deferred-record bodies — module entry refs, var defaults, connection remaps — can be **demanded before that point**, during the entry sweep, and their results are memoized and then re-walked at every consumption site.

Without this flag, an early demand would fold `_build.authConfig` against a still-undefined projection and bake the wrong value into the memo. With it, the walker (`walker.js:1064-1075`) leaves any `_build.*` node whose subtree reads authConfig unevaluated, so the fold happens in the consuming walk, after the projection exists. The guard is conditional on the projection being undefined, so once it exists the flag is a no-op and folds run right here (that's what the comment's last sentence means).

This is what makes `_build.authConfig` resolve inside module entry vars (success tests 98/99; error test B7 covers the position where it stays illegal). Same flag, same reason, at `buildModuleDefs.js:93` — though see finding 8 in the Fable review: the premise doesn't hold for the entry `connections` remap section, which is consumed without a re-walk.

PR #2281, opened 2026-07-20
Fix module page resolver paths in JIT dev builds

Saiby100:
## Problem

A module page backed by a resolver/transformer declares its path relative to the module root, e.g.:

```yaml
pages:
  - _ref:
      resolver: resolvers/makeActionPages.js
```

The full build rebases this path against the module root in the ref walker (step 4), so `lowdefy build` + `lowdefy start` work correctly. But the JIT dev path (`buildPageJit.js`) rebuilds the page refDef directly from the un-rebased authored `_ref` (`pageEntry.resolverOriginal`) and calls `getRefContent` without going through the walker. The relative resolver path then resolves against the app config directory instead of the module root, and navigating to a module resolver page under `lowdefy dev` fails with:

```
[ConfigError] Error importing resolvers/makeActionPages.js.
  Caused by: [Error] Cannot find module '{app}/resolvers/makeActionPages.js'
```

## Fix

Cherry-pick of b790a6587 from the `vite-hono` branch (applies cleanly to `develop`):

- **`rebaseModuleRefPaths.js`** — new shared helper that rebases relative `path`/`resolver`/`transformer` against the module root.
- **`walker.js`** — replace the inline step 4 rebasing block with the helper (identical logic, prevents drift).
- **`buildPageJit.js`** — apply the helper to the page refDef before resolving module page content; keep the JIT-specific refMap update at the call site.
- **`buildPageJit.test.js`** — new test covering a module resolver page: resolver path rebased against the module root, not the app config dir.

## Testing

`packages/build`: all 9 suites / 288 tests pass for `src/build/jit/buildPageJit.test.js` and `src/build/buildRefs/`.

### DIFF
--- a/.changeset/fix-build-jit-module-resolver-path.md
+++ b/.changeset/fix-build-jit-module-resolver-path.md
@@ -1,0 +1,13 @@
+---
+'@lowdefy/build': patch
+---
+
+fix(build): resolve module page resolver/transformer paths against the module root in JIT dev builds.
+
+A module page backed by a `resolver:` (or `transformer:`) declares its path relative to the module
+(e.g. `resolvers/makeActionPages.js`). The full build rebases this against the module root in the ref
+walker, but the JIT dev path (`buildPageJit`) rebuilt the page refDef directly from the un-rebased
+authored `_ref` and called `getRefContent` without going through the walker — so the relative path was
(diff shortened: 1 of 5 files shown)


machielvdw:
0.0.0-experimental-20260720135014

SamTolmay (approved):

PR #2293, opened 2026-07-28
Fix dev-server JIT building-screen strand after login

machielvdw:
## Summary

Investigating the demo-app auth-testing findings (users-fixes F12) showed the dev client can permanently strand a tab on the "Building page…" screen right after a login redirect — a new tab on the same URL always cleared it. The screen is the Suspense fallback around the `/api/page` fetch, and three compounding traps made it permanent: the 401 handler returned a never-settling promise that poisoned the tab's SWR key, the SSE reload recovery signal was a one-shot boolean the engine can only lower after a page mounts (unreachable while suspended), and no page-config fetch had a timeout. This PR makes the dev client un-strandable and adds observability for the trigger.

## Changes

- **@lowdefy/server-dev**: The 401 page-config path now settles with an `authRedirect` value rendered as a redirect screen (with a manual sign-in link as escape hatch) and navigates from an effect. A monotonic reload tick re-renders Routing on every SSE reload event so suspended tabs mint fresh Suspense/SWR keys and recover. All three page-config fetches get `AbortSignal.timeout`, turning stalls into visible ErrorBoundary errors. `RestartingPage` is hoisted above the Suspense boundary so restarts no longer present as "Building page…". Unauthenticated page-config requests are debug-logged.
- **@lowdefy/api**: `resolveAuthentication` debug-logs its two silent unauthenticated exits (session without an active organization, member-row miss) so a 401-after-login is explained by one log line. Membership-wall semantics are unchanged.

## Notes

- The `jitPage.js` 401 response contract (frozen dev-client contract) is unchanged — only the client-side handling of it changed.
- Root-cause analysis lives in the modules-mongodb repo at `designs/users-fixes/06-investigate/findings.md` (F12). The sibling findings F11 (Login without callbackUrl never navigates) and F19 (UpdateSession never refreshes the client session store) are separate framework bugs, deferred to their own passes — this fix de-flakes their verification.
- Verified: full `@lowdefy/api` (782) and `@lowdefy/server-dev` (155) test suites pass, including new log assertions. Live browser checks (blocked redirect, reload-while-suspended recovery, restart visibility, timeout) still need a protected-pages dev app — planned against the modules-mongodb demo on the next experimental snapshot.
- No changeset on this PR by choice — versioning is batched on the auth-upgrade branch.

### DIFF
--- a/packages/api/src/context/resolveAuthentication.js
+++ b/packages/api/src/context/resolveAuthentication.js
@@ -64,6 +64,9 @@ async function resolveAuthentication(context, { auth, headers, strategies }) {
   }
   const activeOrganizationId = session.session.activeOrganizationId;
   if (type.isNone(activeOrganizationId)) {
+    context.logger.debug(
+      `Session for user "${session.user.id}" has no active organization - resolved unauthenticated.`
+    );
     context.user = null;
     return;
   }
@@ -76,6 +79,9 @@ async function resolveAuthentication(context, { auth, headers, strategies }) {
     ],
   });
   if (type.isNone(member)) {
+    context.logger.debug(
+      `User "${session.user.id}" has no member row in organization "${activeOrganizationId}" - resolved unauthenticated.`
+    );
     context.user = null;
     return;
   }
--- a/packages/api/src/context/resolveAuthentication.test.js
+++ b/packages/api/src/context/resolveAuthentication.test.js
@@ -54,12 +54,15 @@ test('sets context.user to null when the session has no active organization', as
   const { auth, findOne } = mockAuth({
     session: { user: { id: 'user_1' }, session: { id: 'sess_1' } },
   });
-  const context = {};
+  const context = { logger: mockLogger() };
 
   await resolveAuthentication(context, { auth, headers: {} });
 
   expect(context.user).toBe(null);
   expect(findOne).not.toHaveBeenCalled();
+  expect(context.logger.debug).toHaveBeenCalledWith(
+    'Session for user "user_1" has no active organization - resolved unauthenticated.'
+  );
 });
 
 test('sets context.user to null when the user holds no member row in the active org', async () => {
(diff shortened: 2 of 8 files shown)


machielvdw:
## Context: how this class of bug relates to the hand-rolled router

When we replaced Next.js we chose a ~60-line custom History wrapper instead of a router library (`designs/vite-hono/routing.md`), reasoning that Lowdefy's routing is trivial: every URL renders the same component with different config. That reasoning was correct — and none of these bugs are in the wrapper itself.

What we underestimated is that a modern router's real job isn't matching URLs. It's managing the **lifecycle of a navigation**: fetch the data for the destination, show a pending state while that's happening, handle a redirect, cancel the fetch if the user navigates somewhere else, deal with errors and timeouts, and refetch when something invalidates. By going custom, we took ownership of all of that — and we rebuilt it by hand out of SWR + Suspense + an SSE listener + module-level counters. Each trap fixed in this PR is a missing piece of that lifecycle:

| Trap in the dev client | What a router with a data layer gives you |
|---|---|
| 401 → `window.location.assign` + a promise that never settles | `throw redirect()` from a loader — the router owns the redirect, cleanly cancels the pending navigation, and can never strand the tab |
| One-shot `reset` boolean + hidden version counters to bust caches | `router.invalidate()` — one monotonic primitive that always re-renders and refetches |
| No abort/timeout on the page-config fetch | Loaders receive an `AbortSignal`; superseded navigations are cancelled automatically |
| Suspense fallback with no escape hatch ("Building page…" forever) | Per-route `pendingComponent` with `pendingMs` and an `errorComponent` |

A tab wedged forever in a pending state that only a new tab clears is a *classic* bug in this territory — it's precisely the failure mode libraries like TanStack Router spent years engineering away.

**Why this PR still patches rather than adopts a router:** the fixes are small and surgical, and swapping the data-loading layer mid-way through the auth-upgrade branch would be a much larger change than the bug justifies. But the lesson is worth recording: the cost of the custom-router decision didn't land in URL matching, it landed in the navigation/data lifecycle. If this machinery keeps sprouting edge cases — or when we build things like link prefetching (already listed as future work in the routing design) — moving the page-config fetch into a loader-based router is the structural fix, because it deletes this fragile layer instead of patching it again.

machielvdw:
0.0.0-experimental-20260728144250

PR #2295, opened 2026-07-29
Resolve Dynamic page content from urlQuery under e2e

### LINKED ISSUE #2294: server-e2e does not pass urlQuery to getPageConfig, so Dynamic blocks never see URL params in e2e
## Summary

`@lowdefy/server-e2e` does not pass `urlQuery` to `getPageConfig`, while `@lowdefy/server` does. A `Dynamic` block whose endpoint reads `_payload: urlQuery.*` therefore works in dev and production but always receives `urlQuery: {}` under `lowdefy build --server e2e`.

The effect is a false negative: a page that works correctly for users fails only when it is tested, so an e2e suite reports a bug that does not exist and cannot cover that page at all.

## The divergence

Both servers scaffold the same two call sites into `.lowdefy/server/src`:

| | `src/html/renderPage.js` | `src/routes/apiPage.js` |
| --- | --- | --- |
| `@lowdefy/server` | `getPageConfig(context, { pageId: resolvedPageId, urlQuery: c.req.query() })` | `getPageConfig(context, { pageId, urlQuery: c.req.query() })` |
| `@lowdefy/server-e2e` | `getPageConfig(context, { pageId: resolvedPageId })` | `getPageConfig(context, { pageId })` |

Both e2e variants have the Hono context `c` in scope and never read its query.

Downstream, `getPageConfig(context, { pageId, urlQuery })` forwards `urlQuery` to `resolveDynamicContent`, which builds the Dynamic block's endpoint payload as:

```js
payload: {
  blockId: block.blockId,
  pageId: shared.pageId,
  params: params ?? {},
  urlQuery: shared.urlQuery ?? {},
}
```

so the omission surfaces as an empty object rather than an error.

Version observed: `0.0.0-experimental-20260707145139` for `@lowdefy/server`, `@lowdefy/server-e2e` and `@lowdefy/api`.

## Reproduction

1. A page with `dynamic: true` containing a `Dynamic` block whose `endpointId` resolves a document by `_payload: urlQuery.someId`.
2. Request it as `/the-page?someId=abc`.

Under `lowdefy dev` / `lowdefy start`, the endpoint receives `someId` and resolves normally. Under `lowdefy build --server e2e` + Playwright, the same request resolves nothing and the block renders its `fallback` slot.

I isolated it by seeding a document, then querying the same filter three ways from the test: directly through the MongoDB driver (matched), through a normal `Api` endpoint on the same page load (matched), and through the `Dynamic` block's resolver (no match). Re-running against a checkout with all of my own changes to that endpoint reverted reproduced it identically, and diffing the generated `.lowdefy/server/src` between a normal build and an `--server e2e` build isolated it to the two lines above.

## Suggested fix

Thread the query through in the `@lowdefy/server-e2e` scaffold exactly as `@lowdefy/server` does:

```js
// src/html/renderPage.js
const pageConfig = await getPageConfig(context, {
  pageId: resolvedPageId,
  urlQuery: c.req.query(),
});

// src/routes/apiPage.js
const pageConfig = await getPageConfig(context, { pageId, urlQuery: c.req.query() });
```

More broadly, this class of bug is invisible because the e2e server is a separate scaffold that can silently drift from the real one in ways that only ever weaken tests. It may be worth either sharing these request-handling modules between the two server packages rather than duplicating them, or adding a test that asserts a `Dynamic` block receives `urlQuery` under the e2e server — the divergence here is precisely in the layer e2e exists to exercise.

## Context

Found while adding Playwright coverage for a report renderer built on a `Dynamic` block (`lowdefy/modules-mongodb`, reporting module). The page reads a report id from the URL, so it is untestable end to end until this is addressed; the assertion is currently parked as `test.fixme`.

JohannMoller:
## Summary

A Dynamic page's endpoint reading `_payload: urlQuery.*` received an empty `urlQuery` object under `lowdefy build --server e2e`, while working correctly in development and production. Pages that function fine for users produced false e2e test failures.

The e2e server scaffold omitted `urlQuery` at both `getPageConfig` call sites. Fixing those surfaced a second gap in the same path: the e2e client never sent a query string on its SPA page-config fetch, and the e2e harness's `setUrlQuery` mutated the URL behind the router's back, so a Dynamic page never re-resolved.

## Changes

- **@lowdefy/server-e2e**: Pass `urlQuery` to `getPageConfig` on the initial HTML render and on the `/api/page/*` route, matching `@lowdefy/server`. The SPA page-config fetch in `client/Page.jsx` now appends `window.location.search` so the route has a query string to read — without this the route fix is inert on client-side navigation. `src/routes/apiPage.js` is now byte-identical to the production server's module, and `client/Page.jsx` differs only by the intentional `stage="e2e"` prop.
- **@lowdefy/e2e-utils**: `ldf.urlQuery(key).do.set(value)` navigates through the app's router instead of calling `history.pushState` directly. A bare `pushState` never notifies the router, so the page config was never re-fetched. `do.set` now also waits for the re-resolved config to arrive, so a following assertion cannot read stale content.

## Notes

- `router.push` was chosen over dispatching a synthetic `popstate` because it is the path a `Link` or `SetUrlQuery` action takes — the harness simulates the app rather than manipulating history directly. It relies on `window.lowdefy` being exposed for `stage === 'e2e'`, which `waitForReady` already depends on.
- `setUrlQuery` throws when the router is absent rather than falling back to `pushState`; a silent fallback would reintroduce the bug invisibly.
- Behaviour change for test authors: `do.set` now triggers a page-config fetch and awaits it. The wait is bounded at 30s and swallows only the timeout, so a missing response cannot hang a suite — the test's own assertion reports the real failure.
- **No tests added.** Neither server package has unit tests for these route modules, and there is no dynamic-page e2e fixture to regression-test against. The client-side half of this fix is exactly the kind of thing that regresses silently, so a fixture app with a Dynamic page reading `urlQuery` is a worthwhile follow-up.
- The module consolidation the issue suggests is not done here. The two `src` trees diverge deliberately in ~8 other files (auth, Sentry, cron, detached endpoints, webhooks, request timeouts), so merging them is a design change rather than a bug fix.

Closes #2294

### DIFF
--- a/.changeset/fix-e2e-utils-set-url-query-router.md
+++ b/.changeset/fix-e2e-utils-set-url-query-router.md
@@ -6,4 +6,4 @@ fix: Set urlQuery through the app router.
 
 `ldf.urlQuery('key').do.set(value)` called `history.pushState` directly, which changed the URL without notifying the Lowdefy router. The page config was never re-fetched, so a Dynamic page kept showing content resolved from the previous query and tests asserting on the update failed.
 
-`do.set` now navigates through the app's router — the same path a `Link` or `SetUrlQuery` action takes — and waits for the re-resolved page config to arrive before resolving, so a following assertion cannot read stale content.
+`do.set` now navigates through the app's router — the same path a `Link` or `SetUrlQuery` action takes. On a Dynamic page it waits for the engine to rebuild the page context from the newly resolved config, so a following assertion or `value()` read cannot see content resolved from the previous query.
--- a/.changeset/fix-e2e-utils-set-url-query-router.md
+++ b/.changeset/fix-e2e-utils-set-url-query-router.md
@@ -1,0 +1,9 @@
+---
+'@lowdefy/e2e-utils': patch
+---
+
+fix: Set urlQuery through the app router.
+
+`ldf.urlQuery('key').do.set(value)` called `history.pushState` directly, which changed the URL without notifying the Lowdefy router. The page config was never re-fetched, so a Dynamic page kept showing content resolved from the previous query and tests asserting on the update failed.
+
+`do.set` now navigates through the app's router — the same path a `Link` or `SetUrlQuery` action takes — and waits for the re-resolved page config to arrive before resolving, so a following assertion cannot read stale content.
--- a/.changeset/fix-server-e2e-page-url-query.md
+++ b/.changeset/fix-server-e2e-page-url-query.md
@@ -1,0 +1,12 @@
+---
+'@lowdefy/server-e2e': patch
+---
+
+fix: Pass urlQuery to getPageConfig in the e2e server.
+
+Dynamic pages built with `lowdefy build --server e2e` received an empty `urlQuery` object, so endpoints reading `_payload: urlQuery.*` resolved no URL parameters. Pages that work in development and production failed only under e2e tests.
+
+The e2e server now resolves dynamic content with the request query string on both paths, matching `@lowdefy/server`:
+
+- The initial HTML render passes `urlQuery` to `getPageConfig`.
+- Client-side navigation forwards the current query string on its `/api/page/*` fetch, and the route reads it.
--- a/packages/servers/server-e2e/client/Page.jsx
+++ b/packages/servers/server-e2e/client/Page.jsx
@@ -47,7 +47,11 @@ function Page({ auth, config, lowdefy }) {
     const unsubscribe = router.subscribe(async ({ pageId }) => {
       const targetPageId = pageId ?? config.rootConfig.home.pageId;
       try {
-        const res = await fetch(`${router.basePath}/api/page/${targetPageId}`);
+        // Forward the current query string so server-side Dynamic block
+        // resolution sees the same urlQuery as an initial HTML load.
+        const res = await fetch(
+          `${router.basePath}/api/page/${targetPageId}${window.location.search}`
+        );
         if (!res.ok) {
           if (targetPageId !== '404') {
             router.replace({ pathname: '/404' });
--- a/packages/servers/server-e2e/src/html/renderPage.js
+++ b/packages/servers/server-e2e/src/html/renderPage.js
@@ -42,7 +42,10 @@ async function renderPage(c, { pageId, status = 200 }) {
     resolvedPageId = home.pageId;
   }
 
-  const pageConfig = await getPageConfig(context, { pageId: resolvedPageId });
+  const pageConfig = await getPageConfig(context, {
+    pageId: resolvedPageId,
+    urlQuery: c.req.query(),
+  });
 
   if (!pageConfig) {
     if (resolvedPageId === '404') {
--- a/packages/servers/server-e2e/src/routes/apiPage.js
+++ b/packages/servers/server-e2e/src/routes/apiPage.js
@@ -23,7 +23,9 @@ import getPathSegments from '../lib/getPathSegments.js';
 async function apiPageHandler(c) {
   const context = c.get('lowdefyContext');
   const pageId = getPathSegments(c, '/api/page/').join('/');
-  const pageConfig = await getPageConfig(context, { pageId });
+  // The client forwards its current query string on the fetch so Dynamic block
+  // resolution sees the same urlQuery as an initial HTML load.
+  const pageConfig = await getPageConfig(context, { pageId, urlQuery: c.req.query() });
   if (!pageConfig) {
     context.logger.info({ event: 'api_page_not_found', pageId });
     return c.json({ pageConfig: null }, 404);
--- a/packages/utils/e2e-utils/README.md
+++ b/packages/utils/e2e-utils/README.md
@@ -151,6 +151,8 @@ await ldf.url().expect.toMatch(/\/tickets\/\d+/);
 
 // Query params
 await ldf.urlQuery('filter').expect.toBe('active');
+// do.set navigates through the app router, so a Dynamic page re-resolves its
+// content for the new query before this resolves.
 await ldf.urlQuery('page').do.set('2');
 
 // Get raw values
--- a/packages/utils/e2e-utils/src/core/url.js
+++ b/packages/utils/e2e-utils/src/core/url.js
@@ -14,7 +14,25 @@
   limitations under the License.
 */
 
+import { type } from '@lowdefy/helpers';
+
 import { waitForReady } from './navigation.js';
+
+// The client router's createUrl prepends basePath, so the pathname handed to it
+// must have basePath stripped or a basePath app receives it twice.
+function createTargetLocation({ basePath = '', href, key, value }) {
+  const url = new URL(href);
+  if (type.isNone(value)) {
+    url.searchParams.delete(key);
+  } else {
+    url.searchParams.set(key, value);
+  }
+  const pathname =
+    basePath && url.pathname.startsWith(basePath)
+      ? url.pathname.slice(basePath.length)
+      : url.pathname;
+  return { pathname, query: url.searchParams.toString() };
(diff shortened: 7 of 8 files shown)


JohannMoller:
## Review

Self-review — I authored this PR, so treat this as a structured note-to-reviewer rather than an independent sign-off.

### Overview

Two related fixes for #2294. `@lowdefy/server-e2e` now passes `urlQuery` to `getPageConfig` on both the initial HTML render and the `/api/page/*` route, and its SPA client sends the query string that the route reads. `@lowdefy/e2e-utils` stops mutating history behind the router's back so a Dynamic page actually re-resolves when a test changes a query param. 7 files, +66/-4.

### What holds up

- **Root cause is fixed at the source**, not patched at the symptom: the omission was in the two call sites plus the client that feeds one of them, and `src/routes/apiPage.js` is now byte-identical to `packages/servers/server/src/routes/apiPage.js`. `client/Page.jsx` differs only by the intentional `stage="e2e"` prop. Future divergence is now visible as a plain diff between the packages.
- **`pageId` parsing is unaffected** by the newly appended query string — `getPathSegments` reads `c.req.path` (`src/lib/getPathSegments.js:21`), which excludes the query in Hono. No `mypage?foo=bar` page ids.
- **Router choice is right.** `router.push` is the path a `Link`/`SetUrlQuery` action takes (`packages/client/src/adapters/createRouter.js:71-90`), so the harness simulates the app instead of forging history entries. It depends on `window.lowdefy`, which `initLowdefyContext.js:80` exposes for `stage === 'e2e'` and which `waitForReady` already relies on — no new coupling.
- **`basePath` handling is correct.** `createUrl` prepends `basePath` (`packages/client/src/adapters/url.js:28`), so stripping it before `push` avoids doubling it on `basePath` apps. `query` as a string is fine — `serializeQuery` passes strings through.
- **No new security surface.** `urlQuery` was already user-controlled input on this path in dev and production. Worth noting that `getPageConfig` deep-copies the cached artifact before resolution, so now that e2e resolves per-query it cannot leak one request's resolved content to another via the config cache.
- License headers are all `2020-2026`; prettier clean; one-function-per-file and single-object-param conventions respected.

### Findings

**1. The wait is weaker than the comment implies (main residual risk).**

`setUrlQuery` awaits `waitForResponse` then `waitForReady`, but for a *query-only* change `waitForReady` is effectively a no-op: it gates on `window.lowdefy.contexts['page:' + pageId]`, and `pageId` is unchanged, so the context already exists and the first poll succeeds. The real guarantee is therefore *"response headers received"* — `res.json()`, `setPageConfig`, and the React commit all still follow.

Consequence: `await ldf.urlQuery('id').do.set('2')` followed by an immediate `.value()` read can still observe pre-navigation content. Retrying `expect.*` assertions hide this; raw `.value()` reads do not.

Cheapest deterministic fix, and it is e2e-only so it costs nothing in production: have `server-e2e/client/Page.jsx` bump a counter after `setPageConfig` (e.g. `window.lowdefy._internal.pageConfigVersion`), and have `setUrlQuery` wait for it to change. Failing that, the limitation should be stated in the README next to the note this PR adds.

**2. `.catch(() => undefined)` swallows more than the timeout it documents.**

```js
.catch(() => undefined)   // comment says "On timeout, fall through"
```

A closed page, a crashed target, or a navigation-destroyed context is discarded the same way, and the test then fails later on an unrelated assertion with the real cause gone. Narrow it:

```js
.catch((error) => {
  if (error.name !== 'TimeoutError') throw error;
})
```

**3. The URL computation belongs in Node, not in `page.evaluate`.**

Everything except `router.push` is pure logic over the current URL, and `page.url()` makes it available in Node — the same pattern already used at `createPageManager.js:257`. Moving it out would:

- make it unit-testable, which matters because `e2e-utils` already has unit tests (`escapeId.test.js`, `resolveTemplateId.test.js`) and this PR adds none;
- let it use `type.isNone` instead of `v === null || v === undefined`, which CLAUDE.md asks for and which is unavailable inside `evaluate` (Playwright serializes the callback, so imports are out of scope) — `@lowdefy/helpers` is already a dependency;
- shrink the browser-side callback to a guard plus one `router.push`.

The `v === null || v === undefined` check is carried over from the original code, so this is not a regression — but the extraction removes the reason for the deviation.

**4. No test coverage.** Neither server package has unit tests for these route modules, and there is no Dynamic-page fixture to regression-test against. The client-side half of this fix is exactly the kind of thing that regresses silently — the server change alone looked complete and was inert on SPA navigation. A fixture app with a Dynamic page reading `urlQuery`, exercised through both an initial load and a `do.set`, would cover both halves. Suggest tracking it as a follow-up rather than blocking here.

### Non-issues checked

- `URLSearchParams` re-serialization (`?flag` → `?flag=`, spaces → `+`) is not a regression: the previous code also mutated `url.searchParams`, which re-serializes the whole query string.
- Overlapping `/api/page/` responses matching the wrong navigation: `do.set` awaits its own response before returning, and `goto` is a full page load that does not issue this fetch, so there is no in-flight request to confuse it.
- No HTTP caching on the e2e HTML response, so query-varying renders cannot be served a stale cached body.

### Verdict

Correct and appropriately scoped; safe to merge as-is. Findings 2 and 3 are small and worth doing in this PR. Finding 1 is the one I would most like a second opinion on — the current behavior is strictly better than before and adequate for assertion-based tests, but it is eventual rather than proven, and I would rather that be a deliberate decision than an accident.

JohannMoller:
## Review findings addressed — aaed25db7

All three findings from the review above are fixed. Finding 1 landed differently from what I proposed there, and the reason is worth reading before reviewing the commit.

### 1. The weak wait — fixed, but not with the counter I suggested

I proposed adding a `pageConfigVersion` counter to `server-e2e/client/Page.jsx`. A better signal already exists: `getContext` replaces `lowdefy.contexts[id]` whenever a new config object arrives for a Dynamic page (`packages/engine/src/getContext.js:88-94`). Comparing config identity across the navigation proves the re-resolved content is in place, and it needs **no client change at all** — so it avoids adding the very divergence between `server/client/Page.jsx` and `server-e2e/client/Page.jsx` that this PR otherwise reduces.

Two constraints shaped the implementation:

- **The rebuild wait is Dynamic-pages-only.** A static page keeps its memoized context (`sameDynamicConfig` is true when `dynamic !== true`), so there is no identity change to observe and the wait would have timed out on *every* `do.set` against a static page — 30s per call.
- **The page-config response wait stays,** because it is the meaningful signal for that static case: blocks reading `_url_query` re-render only once the fetched config is applied. Dropping it would have regressed static-page timing while fixing Dynamic pages.

### 2. Swallowed errors — fixed

The catch now rethrows anything that is not a `TimeoutError`, so a closed page or crashed target surfaces as itself instead of as a later, unrelated assertion failure.

### 3. Logic moved to Node — fixed

`createTargetLocation` is now a pure exported function computing pathname and query from `page.url()`, covered by 12 unit tests. It uses `type.isNone`, so `0` and `''` are now provably preserved rather than incidentally so. `page.evaluate` is reduced to a guard plus one `router.push`.

While writing it I removed a risk I had introduced: the JSHandle was initially passed nested inside an array to `waitForFunction`. Nested-handle serialization is not something I could verify here, so the handle is now the sole top-level argument.

### Still open: finding 4

`@lowdefy/e2e-utils` is at 35/35 tests passing, but the new coverage is only over the pure URL logic. The browser-side behaviour — the router navigation and the context-rebuild wait — is supported by reading the engine and client code, not by execution. A fixture app with a Dynamic page reading `urlQuery`, exercised through both an initial load and a `do.set`, remains the honest way to cover it. I would rather that be tracked as a follow-up than claimed as done here.

Gervwyk (approved):

PR #2296, opened 2026-07-29
Default the post-auth callbackUrl to the app's home page

machielvdw:
## Summary

A successful sign-in with no `callbackUrl` and no `?callbackUrl=` query navigated **nowhere**: `resolveCallbackURL` was a two-rung ladder with nothing at the bottom, and all four navigation sites were written to shrug at `undefined`. The session was minted, the action chain reported success, and the user was left sitting on a live sign-in form with no error anywhere. Reaching a sign-in page without the query param is ordinary rather than exotic — both producers of that param fire only when the user was *bounced* there, so a bookmark, a typed URL, or a link from the signup page produces no destination.

This adds `home: true` as the ladder's bottom rung, adds `callbackUrl: false` so deliberately staying put stays expressible once absence means "go home", and makes an unresolvable destination a legible error instead of a silent return. Implements the `callback-url-default` design (auth-upgrade), the success-path mirror of `error-callback-default`.

It also closes a post-authentication **open redirect** in the query rung the ladder now promotes to a documented feature.

## Changes

- **@lowdefy/client** — `getCallbackUrl`'s `home: true` branch now returns no target when the home config names no page, instead of interpolating the missing value into a URL. `resolveCallbackURL` gains a `callbackUrl === false` short-circuit above the ladder, a `{ home: true }` rung below the query rung, and a throw when nothing resolves. A new `assertCallbackUrlNavigable` rejects `false` on the four paths that forward the value to BetterAuth. `logout` moves onto `resolveTargetURL`, dropping its hand-rolled basePath prefix.
- **@lowdefy/client (security)** — the `?callbackUrl=` rung was guarded by `startsWith('/')`, which accepts `//evil.com` and `/\evil.com`: protocol-relative URLs that navigate off-site. That value never reaches BetterAuth on the email, phone and passkey paths — it goes straight into `window.location.assign` — so `originCheck` / `trustedOrigins` never saw it, and a crafted sign-in link was a post-authentication open redirect with a freshly minted session. Now guarded by `isAppRelativePath`. Pre-existing, fixed here because this PR promotes that rung to a documented ladder rung and rewrites the comment that claimed the protection.
- **@lowdefy/engine** — `createLink` carried the identical unguarded home expression, so the same broken config made a `Link` with `home: true` push `/undefined` into history and write `inputs['page:undefined']`. It now falls through to `noLink`'s `Invalid Link.`. The rule itself moves into a new `getHomePathname`, exported from `@lowdefy/engine` and consumed by the client ladder, so `lowdefy.home` is read in exactly one place instead of being duplicated across two packages.
- **@lowdefy/actions-core** — `callbackUrl` was declared `type: 'string'` on `Login`, `SignUp`, `PasskeySignIn` and `SendVerificationEmail`, which was already wrong: it is a structured `{ home, pageId, url, urlQuery }` target. Corrected to the real shape plus the `false` variant where it is honoured. `newUserCallbackUrl` and `errorCallbackUrl` were mis-typed the same way. `Logout` gained a `params` surface it never had.
- **@lowdefy/build** — three auth fixture snapshots regenerated.
- **@lowdefy/docs** — `actions/Login.yaml`, `actions/Logout.yaml` and `users/login-and-logout.yaml` all stated that an unset `callbackUrl` returns the user to the page the action was called from, which was never true and is now the opposite of the default. Rewritten to document the three-rung precedence and `callbackUrl: false`.

## Key Files

- `packages/client/src/auth/createAuthMethods.js` — the ladder, the `false` opt-out, the throw, the open-redirect guard, and the `logout` fold
- `packages/engine/src/getHomePathname.js` — new; the single reading of `lowdefy.home`
- `packages/engine/src/createLink.js` — the second consumer of that rule
- `packages/client/src/auth/createAuthMethods.test.js` — 92 → 128 tests
- `packages/plugins/actions/actions-core/src/actions/Login/schema.js` — the structured `callbackUrl` shape

## Breaking Changes

- **A sign-in that resolved no destination now navigates to the app's home page** instead of staying put. An app relying on the old no-navigation behaviour — a login form in a modal, an embedded panel, a re-authentication prompt — must now pass `callbackUrl: false`. The previously-only spelling for this was an accident (`callbackUrl: {url: ''}`, which survives as a resolved-but-falsy target); `false` replaces it as the documented surface.
- **`callbackUrl: false` throws** on magic-link, social/OAuth, `SignUp` and `SendVerificationEmail`. There the value is the destination of a hop Lowdefy does not control, so "do not navigate" is unachievable once the browser has left. Both alternatives are worse than an error: omitting the param lands the user on BetterAuth's `"/"` fallback, and silently substituting the home default contradicts an explicit instruction.
- **Magic-link, social and `sendVerificationEmail` under a `basePath` are fixed, not merely reinforced.** BetterAuth resolves its own fallback as `new URL("/", baseURL)` with `baseURL` of `${origin}${basePath}/api/auth`, and a root-relative `/` discards the path — so a subpath-hosted app was dropping those users on the **origin root, outside the app**. They now land on `${basePath}/`.
- **An unresolvable home config now throws** where it previously navigated to `/undefined` (or `/null`) — a 404 with no symptom anywhere.
- **A protocol-relative `?callbackUrl=` is no longer honoured.** Any app that relied on `//host/path` in that query to leave the origin after sign-in must pass an explicit absolute `callbackUrl: {url: ...}` instead, where BetterAuth's origin checks apply.
- **Not a break, deliberately:** a `callbackUrl` that resolves to `null` or to a bare string is treated as absence and falls through the ladder. `null` (an `_if` with no matching branch) previously threw a TypeError. A bare string is the spelling these params were *wrongly declared as* until this PR, so it is common in existing apps: throwing would fail sign-ins that work today, and reading a string as `{ url }` would double-prefix `basePath` on the commonest case of all — a string carrying the already-prefixed query value.

## Notes

- **The throw is reachable, not defensive.** `getHomeAndMenus` genuinely returns `{ configured: false, pageId: null }` for an app with no `homePageId` and no menu link to fall back on, so `home: true` built the literal `"/null"` today. Without the guard in `getCallbackUrl` the new throw would have been dead code.
- **The home default is resolved for the *unauthenticated* caller when `homePageId` is unset.** `getHomeAndMenus` derives `home.pageId` from the menu, and `filterMenuList` filters every `MenuLink` through `authorize` — so on a public sign-in page the fallback is computed from what a signed-out visitor can see. An app whose menu is entirely protected therefore resolves no home and throws, and one with a single public link resolves to that page rather than the user's real home. Setting `homePageId` sidesteps both: the ladder then yields `/` and the server resolves the destination after authentication, when it knows the caller. Worth knowing before relying on the menu fallback.
- **The throw is pre-flight, so "sign-in failed" is the truth.** Every site resolves the destination before its BetterAuth call, so a throw mints no session — a login page mapping the failure to authentication messaging is not lying. A test pins this (`expect(auth.signInEmail).not.toHaveBeenCalled()`); reordering the resolve below the call would turn that message into a lie.
- **`renderPage.js:41` carries a third copy of the same expression** and sends `/` to `${basePath}/null`, which lands on `/404` after a second redirect, logged as `redirect_to_homepage pageId: null`. Left unfixed here — choosing the right status and log line for a home-less app is its own change.
- **Do not "tidy" the query rung through `resolveTargetURL`** for uniformity's sake. Both of its producers (`renderPage.js`, `apiPage.js`) already bake `basePath` into the value they emit, so prefixing there double-prefixes a bounced sign-in to `/app/app/reports`. There is a test pinning this. The same reasoning is why a bare-string `callbackUrl` is not read as `{ url }`.
- **The three regenerated build snapshots were already failing on `auth-upgrade`** before this branch — `newUserCallbackUrl`/`errorCallbackUrl` were added to the Login schema without regenerating them. This clears that backlog alongside the new shape, which is why the snapshot diff is larger than this change alone would produce.
- **`logout` deliberately gains no default** and still reads neither the query nor the home rung; the post-sign-out reload plus the server's page auth fork is the correct landing for a sign-out. It does inherit the `/undefined` correctness fix. The `resolveTargetURL` fold is behaviour-neutral on every input, including the empty-string case where `willNavigate` must stay `false`.
- **No changeset**, matching the sibling `error-callback-default` implementation (`ed27ae72b`) — auth-upgrade features are being rolled up rather than versioned individually.
- **Pre-existing test failures unrelated to this change:** `connection-google-sheets` and `plugin-gcp` fail to load on a patched `buffer-equal-constant-time` via `jwa`. Verified present without these commits. All other packages pass — client 187, engine 384, build 1972, api 782, actions-core 188, server-dev 155.

### DIFF
--- a/packages/build/src/tests/success/35-auth-protected-app/snapshot.json
+++ b/packages/build/src/tests/success/35-auth-protected-app/snapshot.json
@@ -2187,8 +2187,80 @@
             "description": "Send a magic sign-in link to the email address instead of using a password."
           },
           "callbackUrl": {
-            "type": "string",
-            "description": "URL to redirect to after login."
+            "oneOf": [
+              {
+                "type": "object",
+                "description": "Structured callback target for where a successful sign-in lands, basePath-prefixed. Defaults to the app's home page when omitted and no ?callbackUrl= query is present; a ?callbackUrl= query set by the unauthenticated-page redirect wins over that default, and this param wins over both.",
+                "properties": {
+                  "home": {
+                    "type": "boolean",
+                    "description": "Land on the app's home page."
+                  },
+                  "pageId": {
+                    "type": "string",
+                    "description": "The pageId to land on."
+                  },
+                  "url": {
+                    "type": "string",
+                    "description": "The URL to land on. An absolute URL is not basePath-prefixed, so it can be an external landing page."
+                  },
+                  "urlQuery": {
+                    "type": "object",
+                    "description": "The urlQuery to set on the destination."
+                  }
+                }
+              },
+              {
+                "enum": [
+                  false
+                ],
+                "description": "Do not navigate - the session store re-renders the page with the new user in place, for a login form in a modal or an embedded panel. Not valid for magic-link or social/OAuth sign-in, which redirect through an external hop."
+              }
+            ]
+          },
+          "newUserCallbackUrl": {
+            "type": "object",
+            "description": "Structured callback target for where a first-time user lands after a magic link (or social/OAuth sign-in) creates their account. Resolved like callbackUrl - basePath-prefixed. When omitted, BetterAuth defaults it to callbackUrl. Ignored by email and phone sign-in.",
+            "properties": {
+              "home": {
+                "type": "boolean",
+                "description": "Land on the app's home page."
+              },
+              "pageId": {
+                "type": "string",
+                "description": "The pageId to land on."
+              },
+              "url": {
+                "type": "string",
+                "description": "The URL to land on. An absolute URL is not basePath-prefixed, so it can be an external landing page."
+              },
+              "urlQuery": {
+                "type": "object",
+                "description": "The urlQuery to set on the destination."
+              }
+            }
+          },
+          "errorCallbackUrl": {
+            "type": "object",
+            "description": "Structured callback target for where a failed or expired magic link (or social/OAuth sign-in) lands, carrying the ?error= reason. Resolved like callbackUrl - basePath-prefixed. When omitted, defaults to the app's authPages.error page (basePath-prefixed); if authPages.error is unset, BetterAuth's own fallback stands. Ignored by email and phone sign-in.",
+            "properties": {
+              "home": {
+                "type": "boolean",
+                "description": "Land on the app's home page."
+              },
+              "pageId": {
+                "type": "string",
+                "description": "The pageId to land on."
+              },
+              "url": {
+                "type": "string",
+                "description": "The URL to land on. An absolute URL is not basePath-prefixed, so it can be an external landing page."
+              },
+              "urlQuery": {
+                "type": "object",
+                "description": "The urlQuery to set on the destination."
+              }
+            }
           },
           "captchaToken": {
             "type": "string",
@@ -2200,7 +2272,32 @@
     "Logout": {
       "type": "object",
       "params": {
-        "description": "Parameters passed to the logout method."
+        "type": "object",
+        "description": "Parameters passed to the logout method.",
+        "properties": {
+          "callbackUrl": {
+            "type": "object",
+            "description": "Structured callback target for where a signed-out user lands, basePath-prefixed. Unlike the sign-in actions this has no default and does not read the ?callbackUrl= query: with no target the session provider reloads the page and the server re-applies the page auth fork, which sends a protected page to the sign-in page.",
+            "properties": {
+              "home": {
+                "type": "boolean",
+                "description": "Land on the app's home page."
+              },
+              "pageId": {
+                "type": "string",
+                "description": "The pageId to land on."
+              },
+              "url": {
+                "type": "string",
+                "description": "The URL to land on. An absolute URL is not basePath-prefixed, so it can be an external logout landing page."
+              },
+              "urlQuery": {
+                "type": "object",
+                "description": "The urlQuery to set on the destination."
+              }
+            }
+          }
+        }
       }
     },
     "Link": {
--- a/packages/build/src/tests/success/36-auth-roles-app/snapshot.json
+++ b/packages/build/src/tests/success/36-auth-roles-app/snapshot.json
@@ -2123,8 +2123,80 @@
             "description": "Send a magic sign-in link to the email address instead of using a password."
           },
           "callbackUrl": {
-            "type": "string",
-            "description": "URL to redirect to after login."
+            "oneOf": [
+              {
+                "type": "object",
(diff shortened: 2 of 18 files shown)


machielvdw:
### In plain terms

**The bug**

You open the login page directly — from a bookmark, or by typing the URL. You enter your email and password. It works: you're logged in. But the page doesn't move. You're still looking at the login form, with no message, no error, nothing. It looks like the login failed. It didn't.

This happens because Lowdefy asks itself "where should I send this person?" and had exactly two places to look:

1. Did the app config say where to go? (usually no)
2. Was the person sent here from a page they weren't allowed to see? If so, send them back there.

If both answers are "no", the answer was "nowhere" — and nothing complained about it.

The second one only happens when you get *bounced* to the login page. Opening the login page yourself doesn't bounce you, so there's nothing to go back to. That's the whole bug.

**The fix**

Add a third place to look, at the end: **the app's home page.** Every app already knows its home page — the menu uses it, the `/` route uses it, the `Link` action already has a `home: true` option. Now login uses it too.

The order matters, and the new answer goes *last*:

| Where we look | What happens |
|---|---|
| The app config said where to go | Go there (unchanged) |
| You were bounced here from a page | Go back to that page (unchanged) |
| **Neither** | **Go to the home page** ← new |

Putting it last is the whole trick. Nothing that worked before changes. Only the case that was broken gets fixed.

**One extra thing: "please don't move"**

Some login forms *should* stay put — a login popup, or a small login panel inside a page. Those work today by simply not saying where to go. But "didn't say" now means "go home", so those forms would suddenly jump to the home page and lose whatever the person was doing.

So there's a new way to say "stay here on purpose":

```yaml
callbackUrl: false
```

Now "I didn't say" and "I mean nowhere" are two different things, instead of both being blank.

There's one place `false` isn't allowed: login by emailed link, or login through Google/GitHub. In those cases the person leaves the site and comes back, so "don't move" is impossible to honour — the browser has already gone. It says so with an error instead of quietly ignoring you.

**And: stop making up broken links**

If an app has no home page set up at all, the old code built a link to the literal text `/undefined` and sent the browser there — a 404, with nothing anywhere saying why. Two places did this: login, and the `Link` action. Both now say "I can't work out where home is" out loud instead.

**A bonus fix for apps on a subpath**

If your app lives at `example.com/myapp` rather than `example.com`, magic-link and Google logins were sending people to `example.com` — *outside your app entirely*. That's because the fallback used a "/" that threw away the `/myapp` part. Sending the destination explicitly fixes it, so those people now land at `example.com/myapp/`.

**What you'd notice as a user**

| Before | After |
|---|---|
| Log in, nothing happens, looks broken | Log in, land on the home page |
| Bounced to login, log in, return to the page you wanted | Same as before |
| App on a subpath: magic link drops you outside the app | Lands you inside the app |
| App with no home page: silently sent to a 404 | Says what's wrong |
| Login popup that stays put | Same, if you add `callbackUrl: false` |

SamTolmay (approved):

PR #2297, opened 2026-07-29
Add preset shortcuts to the antd date picker blocks

JohannMoller:
## Summary

antd's `DatePicker` and `RangePicker` support a [presets panel](https://ant.design/components/date-picker#date-picker-demo-preset-ranges) of quick-select shortcuts ("Last 7 Days", "This Month"), but Lowdefy exposed no way to configure it. This PR adds a `presets` property to all five antd date picker blocks — `DateRangeSelector`, `DateSelector`, `DateTimeSelector`, `MonthSelector` and `WeekSelector`.

A preset is a `label` (supports html) and a `value`. `DateRangeSelector` takes a `[from, to]` pair; the other blocks take a single date. Values are plain dates — a date string, a timestamp, or a `_date` object — so relative shortcuts compose from the existing operators rather than a new relative-date DSL:

```yaml
- id: report_period
  type: DateRangeSelector
  properties:
    presets:
      - label: Last 7 Days
        value:
          - _dayjs: [now, { subtract: [7, days] }, { format: YYYY-MM-DD }]
          - _dayjs: [now, { format: YYYY-MM-DD }]
      - label: Month to date
        value:
          - _dayjs: [now, { startOf: month }, { format: YYYY-MM-DD }]
          - _dayjs: [now, { format: YYYY-MM-DD }]
      - label: 2026 Q1
        value: ['2026-01-01', '2026-03-31']
```

One side benefit over antd's static preset values: because these are operators, they re-evaluate on every config evaluation, so `_date: now` stays current instead of freezing at page load.

Presets are also gated on `disabledDates`, and getting that right required fixing two pre-existing bugs in `disabledDates` itself — see Notes.

## Changes

- **@lowdefy/blocks-antd**: New `src/getPresets.js` converts config presets into antd's `presets` prop, validates their shape, and narrows or disables presets the calendar cannot select. New `src/schemas/presets.js` adds `datePresets` / `dateTimePresets` / `dateRangePresets` meta schemas (exported from `schemas/index.js`) so docs and property forms generate themselves. New `src/pickerPresetsStyle.css` disables the popup list item of a preset that cannot be selected. `src/disabledDate.js` now compares calendar dates rather than instants, reads `ranges` given as `{ from, to }`, and reports malformed config instead of dropping it. All five picker blocks wire the prop and declare `presets` in their `meta.js`. Each block's `gallery.yaml` gains preset examples, and each `e2e.js` gains `selectPreset` plus preset assertions.
- **@lowdefy/docs**: Gallery sections can set `overflowVisible: true`, so a card does not clip a block popup wider than its column. Set on the two preset sections that were clipped.
- **@lowdefy/docs-content**: Regenerated block docs for the five pickers and for `Calendar`, picking up the new `presets` rows and the corrected `disabledDates.ranges` description.
- **@lowdefy/e2e-utils** / **@lowdefy/block-dev-e2e**: New `getShortcutModifier(page)` resolves a `mod` shortcut from the platform the browser reports, so a test presses the key the app is listening for. Needed to fix the pre-existing e2e failures below.

## Key Files

- `packages/plugins/blocks/blocks-antd/src/getPresets.js` — preset conversion, shape validation, and the narrow-or-disable pass
- `packages/plugins/blocks/blocks-antd/src/disabledDate.js` — wall-clock comparison and `{ from, to }` ranges
- `packages/plugins/blocks/blocks-antd/src/schemas/presets.js` — the meta schemas that drive generated docs and property forms
- `packages/plugins/blocks/blocks-antd/src/blocks/DateTimeSelector/DateTimeSelector.js` — the only block passing `local`, since without `selectUTC` it reads its value as an instant rather than a UTC wall clock
- `packages/docs/templates/blocks/galleryExampleTransformer.js` — the gallery card clipping opt-out

## Notes

**How a preset date is read.** The picker hands a preset value straight to `onChange`, so a preset is read exactly as the block reads and writes its value: the date-only pickers work in a UTC wall clock (`dayjs.utc(value)`), so a fixed date like `2026-01-01` selects the day it names in every timezone. `DateTimeSelector` selects an instant and follows its `selectUTC` setting. Verified byte-identical commits against a calendar-cell click across `Europe/Berlin`, `Pacific/Auckland`, `America/New_York` and `UTC`, for day, month, week and minute granularity.

**A relative date is an instant, not a calendar date — end a `_dayjs` chain with a `format` step.** This is the one thing worth reading before writing a relative preset, and an earlier revision of this PR documented it backwards. On a date-only picker:

| value | result |
| --- | --- |
| `_dayjs: [now, {startOf: month}]` | wrong in every UTC+ zone |
| `_dayjs: [now, utc, {startOf: month}]` | wrong whenever the local and UTC dates differ |
| `_dayjs: [now, {startOf: month}, {format: YYYY-MM-DD}]` | always correct |

`format` short-circuits the chain and returns a plain `YYYY-MM-DD` string, which the picker then reads as a UTC calendar date. Guarded by e2e tests pinned to `Pacific/Auckland` with `page.clock` fixed to 21:30 UTC on 30 June — a moment where the local date is 1 July. Reverting the yaml to `_date: now` fails them with `Expected "2026-07-01" / Received "2026-06-30"`.

**Presets respect `disabledDates`.** Previously a preset the calendar disabled was still offered as a normal shortcut, and clicking it wrote the value into the input, marked the input invalid and left the popup open without committing anything or reporting why — so a `Last 7 days` shortcut next to `disabledDates.min: now` silently did nothing. Now a range is narrowed to the dates it may select, and a preset with nothing to select is listed as disabled:

| config | preset | result |
| --- | --- | --- |
| `min: now` | Last 7 Days `[08, 15]` | `[15, 15]` — start narrowed up |
| `max: now` | Next 7 Days `[15, 22]` | `[15, 15]` — end narrowed down |
| `dates: [08, 15]` | `[08, 15]` | `[09, 14]` — both ends narrowed |
| `min: now` | Last Year `[2025-01-01, 2025-12-31]` | disabled |
| single picker, `min: now` | A Week Ago | disabled (a single date cannot be narrowed) |

The narrowing scans inward from a disabled end, so `min`, `max` and the holes from `dates` / `ranges` are handled uniformly; it only runs when an end is actually disabled and is bounded by the range itself.

**Two pre-existing `disabledDates` bugs had to be fixed for the above to be coherent** — both carry their own changeset, and both are behaviour changes for existing apps:

- `disabledDates` compared instants while reading its own bounds in local time, so the verdict depended on which dayjs mode the date arrived in — and presets arrive UTC-mode. With `min: '2026-07-15'` in Berlin, 14 July was disabled for a calendar cell but allowed for a preset. It also meant a date was disabled in an empty picker but allowed once the block had a value, since the panel then works in the block's UTC wall clock. Bounds and dates are now compared as the calendar dates they name. Berlin and Auckland change; New York and UTC are unaffected.
- `disabledDates.ranges` is documented as a list of `{ from, to }` objects, but only a two-date array was read and anything else was dropped — so a range written the documented way disabled nothing, which is why nobody noticed. Both shapes are now read, and the silent `.filter()` that hid it is gone. The same validation was extended to the rest of the property: a `min`, `max` or `dates` entry that is not a date, and a `dates` or `ranges` value that is not an array, were all ignored silently and now name what is wrong. **An app carrying one of these mistakes will surface an error on the block where it previously rendered a picker that quietly disabled nothing.**

**Rendering a disabled preset.** rc-picker's `PresetPanel` has no disabled state and puts its click and hover handlers on the `li`, giving the block only the label. So the label carries a marker span, and a stylesheet rule disables the `li` via `:has()`. It works because the app's layer order is `theme, base, antd, components, utilities` — a `@layer components` rule outranks antd. The e2e tests assert `pointer-events: none` on the item rather than a class name, so they catch the rule silently not applying, with an enabled preset on the same page as the control.

**Preset shape is validated and throws**, because `dayjs(undefined)` is "now" — a mistyped preset would otherwise render as a plausible shortcut that quietly selects today. A range picker also commits a one-date value as a one-date range, and a single picker throws inside its formatter, so arity is checked at the call site rather than inferred from the configured value. Block properties are never validated against `meta.js` schemas, so this is the only enforcement.

**No changes to the `packages/docs/` block pages** — the user-facing block docs read `meta.js` and `gallery.yaml` through the njk template, so the new property documents itself from the schema. The one `packages/docs` change is the gallery card clipping opt-out: a block renders its popup inside itself, and the card's `overflow: hidden` cut off the range picker's 696px popup in a 524px column, hiding the whole shortcut list. Measured before and after in the browser: the popup moves from 99..795 against a 296..820 clip, to 321..1017 against 296..1360.

## Test status

Full `blocks-antd` e2e suite: **1067 passed, 0 failed**, over two consecutive runs.

That suite had 12 failures when this PR was first pushed, all of which also failed on `origin/vite-hono` with none of this PR's changes (verified in a clean worktree). They are fixed here, along with one test that had been left disabled. None was a product bug — all three causes were in the tests:

- **4 placeholder tests** asserted `Select Date` / `Select Month` / `Select Week` / `Select Date & Time`. There is no Lowdefy placeholder default; the string comes from antd's `en_US` locale, which lowercases the noun. The asserted capitalisations never existed. Note `DateTimeSelector` renders `Select date` — antd has no distinct placeholder for a `showTime` picker.
- **6 shortcut tests** derived `mod` from `process.platform`, but the app resolves it through tinykeys, which reads `navigator.platform` in the browser. Playwright emulates that per project, so a Desktop Chrome project reports `Win32` even on macOS: the app listened for Ctrl while the test pressed Cmd. Linux agrees on Control, so this only broke on macOS hosts. Fixed with a new `getShortcutModifier(page)` in `@lowdefy/e2e-utils` that reads the platform from the page. That also revived `Search › opens search modal when shortcut is pressed`, which carried a `TODO: could not get this to pass` — the same bug, plus a locator scoped to the block when the modal renders in a portal.
- **2 AutoComplete tests** were racing antd: opening the dropdown activates the first option and clears it ~100ms later, so the original `Apple` expectations were correct and the tests were only pressing inside that window. They now wait for the settled state. (Asserting the transient value instead passes in isolation and fails under parallel load — worth knowing if these ever regress.)

**CI does not run these tests.** The `Test` jobs run `pnpm test`, which is jest only; `pnpm e2e` is in no workflow. That is how 12 broken e2e tests went unnoticed, and nothing will catch the next regression either. Out of scope here, but worth a decision.

The e2e app build blocker noted in the earlier revision of this description is resolved by #2300, now merged and merged into this branch.

**Test-run caveat:** the bundled `chromium_headless_shell` binary is broken locally, so these runs used installed Google Chrome via `channel: 'chrome'`. Worth confirming on bundled chromium in CI.

## Not fixed here

- `fullWidth: true` on a docs gallery section is a no-op: it sets `column-span: all`, which the browser ignores because the card's parent is a flex row (`lf-row`). Likely affects the AgGrid docs pages that use it.
- A block popup cannot escape any ancestor with `overflow: hidden`, because the block renders it into a container inside itself for the `popup` cssKey and per-block theming. The docs gallery card was one instance of this; a user's app can hit it too.


JohannMoller:
## Self-review

Reviewed the diff against the antd v6.3.1 / `@rc-component/picker` 1.9.1 source, and re-verified the timezone claims empirically. The core design holds, but there are a few real gaps — including one wrong explanation in the user-facing docs this PR generates. **Fixes to follow in this PR.**

### Verified correct

- **Presets render on week/month pickers.** `PresetPanel` lives in the shared `PickerInput/Popup/index.js` and is picker-type agnostic, so the `MonthSelector` / `WeekSelector` wiring is sound.
- **The core UTC wall-clock conversion holds** across −240 / 0 / +840 offsets: a `_date` object at UTC midnight resolves to the calendar date it names in every timezone.
- **`utc` is a valid `_dayjs` chain method** (`allowedChainMethods`, `operators-dayjs/src/operators/shared/dayjs.js:64`).
- **No double `onChange`.** Worth recording since it looks like a hazard: `triggerSubmit` calls `triggerCalendarChange` *and then* `onChange`, and `DateTimeSelector` wires both `onChange` and `onSelect` to the same handler — so a preset click looks like it should fire the block's `onChange` event twice. It does not. antd only forwards `onSelect` when `picker === 'time'` (`generateSinglePicker.js:124`, `hasLegacyOnSelect`), and rc-picker's `SinglePicker` has no `onSelect` prop at all. Hypothesis refuted.
- Gallery examples are internally coherent, including `dts_presets_utc` correctly pairing its `utc` chain with `selectUTC: true`.

### Findings

**1. The `utc` step guidance is wrong, and it ships into generated docs.** `src/schemas/presets.js:25` tells users to use `_dayjs` "with a utc step, eg. `_dayjs: [now, utc, {subtract: [7, days]}]`".

Verified: for `add` / `subtract` the `utc` step is a **no-op**, because the chain ends in `toISOString()`, which is mode-independent. It is essential for `startOf` / `endOf`. Measured at `TZ=Pacific/Kiritimati` (+840):

| chain | result | final date |
| --- | --- | --- |
| `subtract 7d` **without** `utc` | `2026-07-22T14:19:39Z` | `2026-07-22` |
| `subtract 7d` **with** `utc` | `2026-07-22T14:19:39Z` | `2026-07-22` |
| `startOf month` **without** `utc` | `2026-06-30T10:00:00Z` | **`2026-06-30`** ← wrong month |
| `startOf month` **with** `utc` | `2026-07-01T00:00:00Z` | `2026-07-01` |

So the description illustrates the one case where `utc` makes no difference, and says nothing about the case that silently breaks. Nothing shipped is actually wrong — every gallery example includes `utc` on every chain, including all `startOf` / `endOf` — but the guidance teaches the wrong reason, and a reader could reasonably conclude `utc` is only needed for relative subtraction and then omit it on a `startOf`. The same wording is in the changeset and the PR description.

**2. A missing `value` silently resolves to "now".** `dayjs.utc(undefined)` returns the current time, so `presets: [{ label: 'x' }]` — or a typo such as `values:` — yields a plausible-looking preset that selects today instead of failing. Inconsistently, `value: null` yields `Invalid Date`. Reproduced identically under `America/New_York`, `UTC` and `Pacific/Kiritimati`.

The schema declares `required: ['label', 'value']`, but block properties are **never validated** against `meta.js` schemas — there is no ajv path for them in `packages/build` (`testSchema.js` and `validateAuthConfig.js` are the only consumers), and the metas only feed docs tables and the docs property forms. Per *fix at the source*, the real cause is framework-wide and a guard clause in `getPresets` would be treating the symptom — but silent-and-plausible is strictly worse than the visible `Invalid Date` that `null` already produces, so it deserves a decision rather than silence.

**3. DST spring-forward gap gives a 1-hour error.** `DateTimeSelector` with `selectUTC: true` and a preset whose UTC wall clock lands in the local DST gap. Verified: `2024-03-10T02:30:00Z` under `America/New_York` resolves to `2024-03-10T03:30:00Z`. Inherent to the wall-clock rebuild, very narrow, currently undocumented.

**4. The only branch in the new helper is untested.** `toPickerDate`'s `local` path runs only for `DateTimeSelector`, which has **zero** preset tests — as do `MonthSelector` and `WeekSelector`. All five blocks gained a `selectPreset` e2e helper; three of them have no test that calls it. The 8 new tests cover only the `local: false` path.

**5. Dead metadata.** `docs: { displayType: 'date' }` on the nested `date` schema is never consumed: the `displayType` switch returns early at `case 'yaml'` (`packages/docs/templates/blocks/propertiesFormTransformer.js:256`), so the `case 'array'` recursion into `items` (line 351) never runs for these schemas. Harmless, but it implies a date picker per preset value that never renders.

**6. Nit — unscoped test locator.** `selectPreset` uses page-wide `.ant-picker-dropdown:visible` even though `getPopupContainer` scopes each popup to `#{blockId}_{elementId}_popup`. Fine while only one dropdown is open, but it is a latent strict-mode violation.

### Adjacent pre-existing issue

`DateTimeSelector`'s `onSelect` is dead under antd v6 (see the `hasLegacyOnSelect` note above), so the comment at `DateTimeSelector.js:121` — "we use on select instead of onChange to make the block UX more like the DataSelector which changes date on click and not on ok" — no longer describes reality: that block now only commits on OK. Not caused by this PR, but it sits in a file this PR touches.

### Also still open (pre-existing, flagged in the PR description)

- `e2e/app/lowdefy.yaml` references `TreeSelector/tests/TreeSelector.e2e.yaml`, which does not exist on this branch or on `vite-hono` — blocks the whole blocks-antd e2e suite.
- 4 `renders with default placeholder` failures from antd v6's placeholder casing change (`Select date` vs `Select Date`).

JohannMoller:
## Review fixes applied — 1581780e8

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | Wrong `utc` step guidance in generated docs | Reworded `src/schemas/presets.js` around `startOf` / `endOf`, the case that actually shifts the date, instead of `subtract`, where the step is a no-op. Same correction in the changeset and PR description. Docs regenerated. |
| 2 | Missing `value` silently resolved to "now" | `getPresets` now throws on a missing or unparseable value, with the offending preset as the received value. `undefined` and `null` behave the same way. |
| 3 | DST spring-forward gap | Documented in `getPresets.js` next to the conversion it is inherent to, including why only `DateTimeSelector` with `selectUTC` can show it. |
| 4 | `toPickerDate`'s `local` branch untested | 10 new tests. `DateTimeSelector` gets a `selectUTC` / non-`selectUTC` pair on the same preset, which is what discriminates the branch: noon UTC reads `08:00` local and `12:00` UTC in `America/New_York`. `MonthSelector` and `WeekSelector` cover the month and week snapping. |
| 5 | Dead `docs: { displayType: 'date' }` | Removed. |
| 6 | Unscoped `.ant-picker-dropdown:visible` | New `pickerDropdown` helper scoped to `#bl-{blockId}`, used by both `select` and `selectPreset` in all five helpers. |

One further docs gap found while regenerating: the **`DateRangeSelector` page never carried the date guidance at all**. The docs table does not recurse into the items of a nested array, so `presets.$.value` ended at "The start and end date of the range." — the flagship block for presets documented nothing about how a preset date is read. The date description is now shared from one constant and appended to the range description, so both pages carry it.

### Verification

- All 18 preset tests pass (8 existing + 10 new).
- Full run of the five date picker specs: **96 passed, 4 failed** — the 4 failures are the pre-existing `renders with default placeholder` casing mismatches, unchanged by this PR.
- The `e2e.js` `do` helpers have no consumer in this repo, so the re-scoped locators were verified with a temporary spec driving `select` and `selectPreset` through the helpers for `DateSelector` and `DateRangeSelector` (passing; spec removed afterwards).
- While doing that I hit a timeout calling `select()` on a range picker that already has a value. **Not a regression** — it reproduces identically with the old page-wide locator, so it is a pre-existing limitation of that helper.

### Still open (pre-existing, not addressed here)

- `e2e/app/lowdefy.yaml` references `TreeSelector/tests/TreeSelector.e2e.yaml`, which does not exist on this branch or on `vite-hono`. The suite runs only with that `_ref` commented out.
- The 4 antd v6 placeholder-casing test failures (`Select date` vs `Select Date`).
- `DateTimeSelector`'s `onSelect` prop is dead under antd v6, so the comment at `DateTimeSelector.js:121` no longer describes the block's behaviour.

All three are separate concerns from presets, so they are left for their own change rather than widened into this diff.

Gervwyk (approved):

PR #2298, opened 2026-07-29
UpdateSession must actually refresh the BetterAuth client session store

machielvdw:
## Summary

`UpdateSession` is documented to refresh two things: the BetterAuth client session store, and `_user`'s roles and attributes. It only did the second. Its `getSession` call was a plain path request that never reached the session atom (`/get-session` is not in BetterAuth's `atomListeners` matcher), and its result was discarded — so every session-user field (`profile`, `name`, `emailVerified`, deployment `additionalFields`) stayed stale on the client until a full page load. Reproduced as an onboarding page that saves a profile, fires `UpdateSession`, then links to a router page — which read the stale marker and bounced the user straight back to the form they had just submitted.

Implements the `update-session-store-refresh` design (auth-upgrade). Three further defects surfaced in the same code ship alongside: `UpdateSession` could silently null the caller, an impersonation switch could publish a caller with no roles, and the client merge dropped two fields the server-resolved caller carries.

## Changes

- **@lowdefy/client**: `updateSession` replaces the discarded `getSession` call with an awaited `auth.refreshSession({ disableCookieCache: true })` — the session atom's own refetch, which is the store's only writer and resolves only after the store is written. A failed refresh now throws to the action's `catch` instead of reporting success over a stale store, and a session that is present while `/api/user` resolves no user throws instead of silently nulling `lowdefy.user` (that is the admission wall rejecting a real session, not a logout).
- **@lowdefy/server, @lowdefy/server-dev** (templates kept byte-identical): the auth wrapper gains `refreshSession`, returning the session atom's `{ data, error }` so `unwrap` works and failures surface. A `sessionScoped` helper merges `fetchOptions: { disableSignal: true }` into the ten methods whose endpoints flip `$sessionSignal` while mutating an existing session (`setActiveOrganization`, `acceptInvitation`, `leaveOrganization`, `removeMember`, `impersonateUser`, `stopImpersonating`, `changePassword`, `revokeOtherSessions`, `twoFactorEnable`, `twoFactorDisable`) — BetterAuth's own signal-driven refetch fires 10ms after those endpoints and aborts the refetch `UpdateSession` awaits, leaving it resolved with nothing written. Calls that can *establish* a session (the sign-ins, and the two-factor, phone-number and passkey verifies) keep their signal: its refetch is what makes the store notice a login when no navigation follows. `UpdateSession` is now the only refresher of an established session. The `Session` merge renders the last resolved caller on an id mismatch (an impersonation switch mid-refresh no longer publishes `roles: []` to role-guarded pages) and carries `activeOrganizationId` and `impersonatedBy` from the resolved caller — both live on the session row, not the session user, so the client dropped them the moment the first session fetch settled.

## Breaking Changes

- **`UpdateSession` now surfaces failures it previously swallowed.** A failed session refresh, or a session that exists while the server resolves no user, throws to the action's `catch`/`onError` — previously the first proceeded silently over a stale store and the second silently nulled `_user` (a phantom client-side logout).
- **BetterAuth's background session refetch no longer fires after the ten session-changing actions above.** Chain `UpdateSession` after them — which is already the documented convention, and now actually guarantees a fresh `_user` when the next action runs.

## Notes

- Stacked on #2296 (`feat/callback-url-default`) — same two client files; retarget to `auth-upgrade` when it merges.
- Two test suites in `server-dev/lib/client/auth/` guard the templates: `authTemplateSync.test.mjs` keeps the server and server-dev copies byte-identical (a divergence names the file), and `betterAuthSessionContract.test.mjs` pins the four undocumented better-auth behaviours this PR relies on — the session atom's awaitable `refetch`, the non-writing `getSession` path call, the never-rejecting failed refetch, and the signal-driven background refetch that `disableSignal` suppresses, including the competing-fetch abort race. A better-auth upgrade that changes any of these now fails in CI instead of visibly only in a running app.
- The end-to-end behaviours remain the design's Gate, run in a live app against **both** templates: exactly one `/get-session` (carrying `disableCookieCache`) after `SetActiveOrganization`/`ImpersonateUser`/`StopImpersonating`; `_user.activeOrganizationId` surviving the first session fetch; `_user.roles` never transiently emptying during an impersonation switch; the onboarding-bounce reproduction closing.
- Verified against installed `better-auth@1.6.23`: `proxy.mjs:49` honours `options.disableSignal`, the session atom's value exposes awaitable `refetch`, and `authClient.$store.atoms.session` is the real atom.
- `disableSignal` also silences the organization plugin's org atoms for those calls — none are exposed to app config today; a future design exposing an org hook must refresh it through its own action (recorded in design Decision 6).
- No changeset, matching the sibling auth-upgrade features (rolled up rather than versioned individually).
- No docs change needed: the docs already describe the behaviour this PR makes true.
- Design: `lowdefy-design/designs/auth-upgrade/features/update-session-store-refresh/design.md` (twice-reviewed, all findings resolved).

### DIFF
--- a/packages/client/src/auth/createAuthMethods.js
+++ b/packages/client/src/auth/createAuthMethods.js
@@ -399,13 +399,20 @@ function createAuthMethods(lowdefy, auth) {
     return unwrap(auth.stopImpersonating());
   }
 
-  // Bypasses the cookie cache (a live re-resolve), so role, attribute or
-  // session changes surface immediately instead of after cookieCache.maxAge.
-  // Roles and merged attributes come from the server-resolved caller - the
-  // active member row read the base session does not carry.
+  // Refreshes the BetterAuth client session store through an awaited
+  // store refetch, bypassing the cookie cache (a live re-resolve) so role,
+  // attribute or session changes surface immediately instead of after
+  // cookieCache.maxAge. The store holds a fresh session user before this
+  // resolves. Roles and merged attributes come from the server-resolved
+  // caller - the active member row read the base session does not carry.
   async function updateSession() {
-    await unwrap(auth.getSession({ disableCookieCache: true }));
+    const session = await unwrap(auth.refreshSession({ disableCookieCache: true }));
     const { user } = await auth.getResolvedUser();
+    if (session && type.isNone(user)) {
+      // A session with no resolved caller is the admission wall rejecting a
+      // real session, not a logout - surface it instead of nulling the caller.
+      throw new Error('UpdateSession failed: a session is active but the server resolved no user.');
+    }
     if (auth.updateResolvedUser) {
       auth.updateResolvedUser(user ?? null);
     }
--- a/packages/client/src/auth/createAuthMethods.test.js
+++ b/packages/client/src/auth/createAuthMethods.test.js
@@ -34,6 +34,9 @@ function setup({ signInResult, signUpResult } = {}) {
       Promise.resolve({ data: { token: null, user: {} }, error: null })
     ),
     deletePasskey: jest.fn(() => Promise.resolve({ data: { status: true }, error: null })),
+    getResolvedUser: jest.fn(() =>
+      Promise.resolve({ user: { id: 'user-1', roles: ['admin'], attributes: {} } })
+    ),
     getSession: jest.fn(() =>
       Promise.resolve({
         data: { session: { activeOrganizationId: 'org-1' }, user: {} },
@@ -52,6 +55,9 @@ function setup({ signInResult, signUpResult } = {}) {
     phoneNumberSendOtp: jest.fn(() => Promise.resolve({ data: { status: true }, error: null })),
     phoneNumberVerify: jest.fn(() =>
       Promise.resolve({ data: { token: 't', user: {} }, error: null })
+    ),
+    refreshSession: jest.fn(() =>
+      Promise.resolve({ data: { session: {}, user: { id: 'user-1' } }, error: null })
     ),
     removeMember: jest.fn(() => Promise.resolve({ data: { member: {} }, error: null })),
     requestPasswordReset: jest.fn(() => Promise.resolve({ data: { status: true }, error: null })),
@@ -88,6 +94,7 @@ function setup({ signInResult, signUpResult } = {}) {
     updateOrganization: jest.fn(() =>
       Promise.resolve({ data: { id: 'org-1', name: 'Acme Inc' }, error: null })
     ),
+    updateResolvedUser: jest.fn(),
   };
   return { auth, lowdefy, assign };
 }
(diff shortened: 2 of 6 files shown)


machielvdw:
**This PR in simple language**

**What was broken.** The `UpdateSession` action is supposed to make the browser's copy of the logged-in user fresh. It's documented to refresh two things: BetterAuth's client-side session store (the user's name, profile, email-verified flag, custom fields) and `_user`'s roles and attributes. It only did the second. The call meant to refresh the store didn't actually write anything into it, and its result was thrown away. So after changing user data and calling `UpdateSession`, pages kept seeing the old user until a full page reload. Real-world symptom: an onboarding page saves your profile, calls `UpdateSession`, sends you onward — and the next page still sees "profile not filled in" and bounces you straight back to the form you just submitted.

**The fix.** `UpdateSession` now uses BetterAuth's own store-refresh mechanism, actually waits for the store to be written before moving on, and skips the cookie cache so it gets live data instead of a cached copy.

**A timing problem that had to be solved with it.** After certain actions (switching organization, impersonating a user, changing password, …), BetterAuth fires its *own* background session refresh a few milliseconds later. That background refresh cancels the one `UpdateSession` is waiting on — leaving it "finished" but with nothing updated. So for the ten actions that change the current session, that automatic background refresh is now turned off, and chaining `UpdateSession` after them (already the documented pattern) is what refreshes the client. Sign-in and verification actions keep their automatic refresh, because that's how the client notices you've logged in.

**Three more bugs fixed in the same code:**
- If the refresh fails, `UpdateSession` now throws an error you can catch, instead of pretending it worked.
- If a session exists but the server refuses to resolve a user for it, `UpdateSession` throws instead of quietly making the client look logged out.
- `_user.activeOrganizationId` and `_user.impersonatedBy` no longer vanish a moment after the page loads, and switching impersonation no longer briefly shows a user with no roles to role-guarded pages.

**What existing apps should know (breaking):**
- `UpdateSession` can now fail loudly (throw to `catch`/`onError`) where it used to fail silently.
- After the ten session-changing actions, the client no longer refreshes itself in the background — chain `UpdateSession` after them to see the change.

**Tests.** Unit tests cover the new `UpdateSession` behaviour. Two new test suites guard the parts that live in the server templates: one keeps the `server` and `server-dev` copies byte-for-byte identical (they're synced by hand), and one pins the undocumented BetterAuth internals this fix relies on, so a future BetterAuth upgrade that changes them fails in CI instead of silently breaking `UpdateSession` in a running app.

---

*Review follow-up (commits 6a1a619, 6756473): a code review verified every claim against the installed BetterAuth package and found `TwoFactorEnable`/`TwoFactorDisable` also trigger the background refresh — they're now included in the ten suppressed actions — and the two test suites above were committed. All 296 tests pass.*

PR #2304, opened 2026-08-06
feat: MCP endpoints-as-tools + auth.strategies on the vite-hono line

Yianni99:
Ports the MCP capability from #2279 plus `auth.strategies` from `auth-upgrade` onto the vite-hono line, so apps pinned to vite-hono experimental releases can serve MCP tools without moving to the in-flight auth-upgrade base.

## Why a port

Production-bound apps pin releases cut from `vite-hono`; `auth-upgrade` is ~294 commits ahead and still moving, and re-pinning an app onto it just to get MCP would make that app the auth rework's QA surface. The ported `mcp:` and `auth.strategies` config blocks are **byte-identical** to auth-upgrade's (verified by diff), so migrating back later is a re-pin + re-test, not a config rewrite. This branch retires once auth-upgrade stabilizes and lands.

(#2279 itself has separately been brought current with `auth-upgrade` — merge + green suites available locally whenever it should land upstream.)

## What it carries

**MCP endpoints-as-tools** (from #2279, endpoints-only)
- Root `mcp:` config block (`name`, `version`, `endpoints`) → `POST /api/mcp` (stateless streamable HTTP, `@hono/mcp` + `@modelcontextprotocol/sdk`) in both servers.
- Endpoint `description`/`payloadSchema` become tool description/inputSchema; `payloadSchema` is passed through a clean-build-artifact step so `~arr`/`~k` markers never reach clients.
- `tools/list` filters by the caller's authorization; `tools/call` re-authorizes inside `callEndpoint`.
- Agent tools deliberately not ported (vite-hono agent configs carry no `auth` entity); `mcp.agents` is rejected by schema.

**`auth.strategies`** (from auth-upgrade — apiKey/jwt header auth)
- Build validation/defaults, verifiers hosted in `plugin-next-auth` (source hosts them in `plugin-better-auth`, which doesn't exist here), `resolveStrategyCaller` yielding `apiKey:{strategyId}:{keyId}` callers with the strategy's `roles`.
- Server glue: no session → try strategies → synthesize `session.user`, so `createAuthorize`/`_user` work unchanged. Apps without strategies short-circuit (no strategies resolved, no behaviour change from that path).
- `AuthenticationError` + 401 mapping: unauthenticated calls to role-gated endpoints surface as 401 instead of the masked "does not exist". **Note for review — this semantic change applies to every app, strategies or not** (matching auth-upgrade's semantics): an anonymous caller can now distinguish "gated" from "nonexistent" by endpoint name, in exchange for correct 401 behaviour for API clients. Wrong-role callers stay masked. Flagging explicitly for sign-off.

## Verification

- Suites green: `@lowdefy/build` (1749), `@lowdefy/api` (496, incl. an apiKey→MCP integration test), `@lowdefy/plugin-next-auth` (25), `@lowdefy/server-dev` (171), `@lowdefy/errors` (130).
- Scratch-app smoke: anonymous `tools/list` → public tools only; keyed → gated tools appear; `tools/call` carries `_user.id = apiKey:{strategyId}:{keyId}` + roles into the routine; bad/missing key → 401-shaped MCP error; strategies-free app unchanged.
- **Verified against a real production-bound app end-to-end** (local CLI + `--server-directory` against this branch): 15 role-gated endpoints served as MCP tools behind per-author API keys — a full multi-step tool-call workflow (creates, cross-document writes, validation, status flips, deletes) over `POST /api/mcp`, clean inputSchemas, structured error-as-data responses, and header-auth identity flowing into every write.

## Ask

An **experimental release cut from this branch** (`pnpm release:experimental`) so the downstream app can bump its pin.

## Notes for reviewers

- Snapshot churn decomposes into the new `mcp.json` / `plugins/auth/strategies.js` artifacts and the resulting keyMap renumbering; the full snapshot suite passes on a fresh regen.
- Pre-existing issues observed on the experimental line while smoke-testing (not addressed here): plain `/api/endpoints/*` reject responses expose `~e` stack traces; endpoint responses can leak `~k`/`~d` serializer artifacts (the MCP path is clean — cleaned schemas and results verified).


Gervwyk:
Updated: merged current `vite-hono` (incl. develop) and ported the MCP serverInfo branding (`mcp.title`, `mcp.websiteUrl`, `mcp.icons`, with build-artifact markers stripped from icons) from auth-upgrade commits 6f986374a / 81f8ba7ed, adapted to this branch's string endpoint entries.

Gervwyk:
Closed the pre-auth existence oracle flagged in the description before merging (6d38790be): on an auth-configured app a session-less caller now gets the identical `Authentication required for API endpoint "..."` answer for a missing id, an `InternalApi` id and a protected id — on `/api/endpoints/*` and MCP `tools/call` (unknown tool included). Authenticated callers, apps without auth and system runs keep the opaque does-not-exist. Mirrors auth-upgrade's 72e018386 with the guard reading the build's `auth.json` on the miss path instead of `authEnforcement` plumbing. api 513 / server-dev 171 / build mcp+strategies green. Also merged current `vite-hono` and ported the MCP serverInfo branding.

PR #2312, opened 2026-08-18
feat: Static reports as an opt-in plugin (@lowdefy/plugin-reports)

JohannMoller:
## Static Reports — plugin architecture

Adds a static, shareable (PDF / Excel) rendering of a Lowdefy page that **reuses the page's existing block config and requests** instead of re-authoring the dashboard as a document. Ships as one opt-in plugin, `@lowdefy/plugin-reports`, reached through the platform's ordinary request seam, leaving core with four minimal, gated changes.

Implements the design at `lowdefy-design/designs/static-reports/design.md` and the approved plan. Built on `vite-hono` (its base), so this PR targets `vite-hono`.

### What's in it

**Plugin — `@lowdefy/plugin-reports`** (self-contained render pipeline)
- IR (closed node set, v2) → pdfmake and exceljs renderers; `walkBlocks` drives the page's engine-evaluated block tree through per-block renderers.
- `Reports` connection, `RenderReport` request (`meta.appAccess: true`, base64 envelope), `DownloadFile` action.
- Headless page evaluation reusing the engine; `_user`-guard so a userless (scheduled) run refuses to render a user-scoped page rather than emitting a plausible-but-wrong document.
- Image resolution via `origin` + guarded fetch (SSRF guardrails: private/loopback DNS refusal, redirect:error, timeout, size cap, `image/*`), own-origin exempt. No disk reads (public/ isn't on disk serverless).

**Core changes (all gated on `@lowdefy/plugin-reports` in `plugins:`)**
- `packages/api`: opt-in `app` capability passed to resolvers whose `meta.appAccess === true` (render-depth capped at 1), wrapping the page-request `callRequest` and `getPageConfig`.
- `packages/build`: the `report` block schema key; `writeBlockStaticImports` → `build/plugins/blocksStatic.js` and `writeReportStyles` → `build/reports/styles.css`.
- servers: `apiContext` injects the new build artifacts.

**Block-package renderers** (`[Block].static.js` + `static: true` meta + `./static` export)
- `blocks-echarts` (EChart → svg via SSR), `blocks-aggrid` (grids → grid), `blocks-antd` (10 renderers), `blocks-markdown`, `blocks-basic` (Img/Icon/Html via takumi).

### Out of scope
Chromium capture (design §9) is a separate sub-design.

### Testing
- Per-package static renderer suites green; plugin unit suites (IR, walkBlocks, pdf/xlsx, evaluatePage, generateReport, resolveImage) green.
- Verified end-to-end against a live dev server with a test app (gitignored): plugin links via `workspace:*`, both gated artifacts generate, `blocksStatic.js` wires 13 renderers, and Download PDF / Download Excel produce files.


JohannMoller:
## Review

The architecture is reasonable and the core seam is small, but the plugin ships several silent-wrong-output paths, one crash in existing dynamic-block code, a native binary added to every Lowdefy app, and no docs or changeset. Not ready to merge as-is.

### Blocking

- **Failed `onInit` requests produce a plausible but empty report.** The headless engine wires `handleError` to `logger.error` only (`createHeadlessLowdefy.js:251`). The engine catches every action error and resolves, so a failing `Request` in `onInit` leaves `_request` as `null` and the PDF ships with empty tables. In the browser the user sees an error toast; on a schedule the wrong document goes out. Collect action errors on the handle and fail the render.
- **Dynamic blocks crash on `report.sheetName`.** `validateReport` pushes into `pageContext.sheetNameRefs`, but the pageContext built in `buildDynamicBlocks.js:75-118` has no such array. Any Dynamic endpoint returning a block with a `sheetName` throws a raw TypeError at runtime. Share a pageContext factory with `buildPage` so new ref arrays cannot be forgotten.
- **`@takumi-rs/core` is a hard dependency of `blocks-basic`.** A napi Rust binary with eight platform packages, now installed by every Lowdefy app for the Html renderer alone. This contradicts the PR's own "an app that never declares the plugin pays nothing". Move it to the plugin and hand the engine to the renderer through `context`, the way `icons` and `fonts` already arrive.
- **XLSX formula injection through object cell values.** `toXlsx.js:162-179` pushes `cell.value` straight into ExcelJS, which types by shape. A row value of `{ formula: 'HYPERLINK(...)' }` from an AgGrid `field` becomes a live formula. Restrict cell values to primitives and `Date` in the IR validator.
- **Flex-child tiles overflow the page.** `walkBlocks.js:294` renders every `flex: 1` child at the full available width, and EChart and Html use that width literally. Three flex charts become three full-width SVGs in three columns. Defer rendering flex children until the row closes and divide the width.
- **`config.requestTimeout: 0` becomes a one-second deadline.** The schema documents zero as "disable". `RenderReport.js:96` computes `max(0 - 2000, 1000)`. Treat zero as no HTTP deadline and fall back to the plugin default.
- **The `_user` guard keys on the wrong signal.** `invocation: app.user ? 'user' : 'system'` means an anonymous visitor on a public page that reads `_user` gets the "cannot render on a schedule" error. The real system marker is `createAuthorize({ system: true })`. Expose an explicit system flag through `createApp`.

### Design and correctness

- **`walkBlocks` flattens all slots and list items into one sibling list.** Areas are emitted in config key order rather than component order, list items pair up on rows the browser never shows, and Tabs/Collapse print every panel title first and all content after. `linearizePanels` documents this as acceptable loss; it is a misleading document. Change the renderer contract to children grouped by slot before more renderers are written on the flat one.
- **The IR is not actually closed.** `nodes.js` validates `kind` and row widths only. Empty headers, ragged rows, and non-string SVG pass validation and crash pdfmake outside the per-block isolation, taking the whole report down with pdfmake's internal message and no blockId.
- **Servers now import icons and client operators for every app, ungated.** All three `apiContext` files statically import `icons.js`, `operators/client.js`, and `clientJsMap.js` into every request context. Only `blocksStatic.js` is gated. Emit one gated `reportsRuntime.js` artifact and import that.
- **`app.readConfigFile` hands out the raw build reader.** `createApp.js:47` exposes every `pages/*.json` and `connections/*.json` to any resolver with `appAccess`, bypassing `authorize`. The comment claims "accessors, not gates". Expose narrow readers for block metas, global, and the stylesheet.
- **Own-origin SSRF exemption trusts the Host header.** `origin` comes from `c.req.url`, and relative image paths skip the private-network guard against it. Take origin from config or drop the exemption. The DNS-rebinding gap the code calls unfixable is fixable with an undici `Agent` custom `lookup`. CGNAT (`100.64.0.0/10`) and multicast ranges are also unhandled.
- **`report.rendering` is schema-valid and always rejected.** The schema lists it as a valid key and `validateReport` throws for any value. The docs endpoint will advertise a key that never works. Drop it until it exists.
- **Reserved keys apply everywhere but are read in one place each.** `size`, `orientation`, `title`, `header`, `footer` are only read from the page block; `sheetName` and `exclude` only from leaves. Nothing warns when they land elsewhere, and nothing warns when a page uses `report:` without the plugin declared.
- **`reportOptions[block.id]` is dead.** Engine ids are the qualified `prefix:blockId:n` form and options are keyed by `blockId`, so the third lookup in `walkBlocks.js:144` can never match.
- **Xlsx sheet names still throw.** ExcelJS rejects leading/trailing apostrophes and the name `History`; the build pattern accepts both.
- **Renderer divergences from their React blocks.** Title defaults to level 4 where the block defaults to 1. Descriptions swaps label precedence and drops `title`. HTML in `content` renders as literal tags. EChart ignores `theme` and string heights. AgGrid ignores `defaultColDef` and column groups.
- **Generation timeout excludes queue wait.** `acquire()` is awaited before `withTimeout` starts, so under load a render continues after the client has been answered.
- **`onMount` never runs and nothing says so.** A page loading data in `onMount` renders empty with no warning.
- **813 KB Roboto base64 blob duplicated.** `fonts/robotoBase64.js` is the same four faces already in `actions-pdf-make`, and pdfmake ships them too.

### Standards and hygiene

- **Nested ternaries** in `writeReportStyles.js:68`, `walkBlocks.js:211`, and `EChart.static.js`. `== null` in `Divider.static.js` and `walkBlocks.js`. Hand-rolled null checks in `Markdown.static.js`. `typeof x !== 'string'` throughout instead of `type.isString`.
- **One function per file** is ignored across the plugin: `toPdfMake.js` (596 lines), `markdownToPdfMake.js` (467), `walkBlocks.js` (356), `resolveImage.js` (340), `generateReport.js` with eight functions and a module-level semaphore.
- **Copy-pasted helpers.** `isBlank` and `cell` exist in five packages. `DEFAULT_CONTENT_WIDTH` is defined four times. Put them in `@lowdefy/block-utils` or a small IR package.
- **Comments reference a design doc that is not in this repo** ("design §5", "core change 2, point 6", "task 17"). Several are already stale, e.g. the note in `createApp.js` saying the artifacts are "undefined until the reports build steps land".
- **Tests mock the thing under test.** `createApp.test.js` mocks `callRequest` and `getPageConfig`. `RenderReport.test.js` mocks `generateReport`. No test shows an unauthorised session getting `null`, or a nested RenderReport being refused. `moduleGraph.static.test.js` hardcodes type lists instead of deriving them from `metas.js`, so it cannot catch a `static: true` meta with no renderer. `Html.static.test.js` imports fonts from another package's `src/` by relative path.
- **No changeset, no user docs, no code-docs.** A new package plus changes to build, api, three servers, and five block packages ships with nothing under `.changeset`, `packages/docs`, or `code-docs`. The `report` block key, `Reports` connection, `RenderReport`, and `DownloadFile` are undocumented.
- Unrelated prettier reformatting hunks in `lowdefySchema.js` and `buildPage.js` add noise.

### What is fine

Request authorisation is preserved: `app.callRequest` re-enters `callRequest`, which runs `authorizeRequest` against the invoking session, and `getPageConfig` returns `null` for both unknown and unauthorised pages. The render-depth cap works. Moving `@lowdefy/operators` to devDependencies is correct since the plugin has no runtime import of it. The `objectToThemeVars` extraction is behaviour-preserving. The 91 snapshot changes are only the new empty `blocksStatic.js` line.

### Recommendation

Split into two PRs. Land the core seam first (gated runtime artifact, narrow `app` accessors, dynamic pageContext fix, changeset). Then land the plugin once the walker contract groups children by slot, the IR validator is actually closed, action errors fail the render, and takumi lives in the plugin.

packages/build/src/lowdefySchema.js
```diff
@@ -1253,6 +1253,59 @@ export default {
             type: 'Block "areas" should be an object.',
           },
         },
+        report: {
```
SamTolmay, line 1285: This should be print, not report

JohannMoller: I'm not so sure about print

machielvdw:
0.0.0-experimental-20260909123240

machielvdw:
0.0.0-experimental-20260911113654

machielvdw:
0.0.0-experimental-20260911131031

PR #2318, opened 2026-08-21
fix: `hasFeedback: false` leads to strange display

### LINKED ISSUE #1471: `hasFeedback: false` leads to strange display
#### Describe the bug

If I use `hasFeedback: false` in input block type `DateSelector`, feedback icon is still displayed, but outside of input field and in gray (see screenshot). This occurs after `Validate` action is executed. I don't know how it should be displayed properly, but I would expect, that there is no feedback icon anymore (maybe also no feedback text).

This may also affect other input block types.

#### To Reproduce

```yaml
id: test
type: PageHeaderMenu
properties:
  title: Test
blocks:
  - id: my_box
    type: Box
    layout:
      size: 400
    blocks:
      - id: my_date_selector
        type: DateSelector
        properties:
          label:
            hasFeedback: false
        validate:
          - message: Not passed
            status: error
            pass: false
      - id: my_validation_button
        type: Button
        events:
          onClick:
            - id: my_validation
              type: Validate
```

#### Screenshots

image

#### Lowdefy Version

3.23.3

Roshan931:
Fixes #1471

## Root cause

label.hasFeedback property is never read by Label block logic

## Changes

- labelLogic now derives `hasFeedback` from `properties.hasFeedback` and gates both the feedback message (`showFeedback`) and a new `showFeedbackIcon` flag on it; Label.js renders the status icon only when `showFeedbackIcon` is true. Validation itself and the input's error/warning styling are untouched.

### DIFF
--- a/packages/plugins/blocks/blocks-antd/src/blocks/Label/Label.js
+++ b/packages/plugins/blocks/blocks-antd/src/blocks/Label/Label.js
@@ -61,6 +61,7 @@ const Label = ({
     rowStyle,
     showExtra,
     showFeedback,
+    showFeedbackIcon,
     wrapperCol,
   } = labelLogic({ blockId, blockClassNames, content, properties, required, styles, validation });
   if (!iconMap) {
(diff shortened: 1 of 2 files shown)


Gervwyk:
Thank you — applied to `develop` via #2338 (cherry-picked with your authorship, plus a changeset), from where it flows into the v6 and v7 release lines. Closing this one since we release from `develop` rather than `main`.

PR #2320, opened 2026-08-21
fix: Visible change removes value in list from state

### LINKED ISSUE #1120: Visible change removes value in list from state
#### Describe the bug

If a list is placed inside a box, and the box visiblity is set to false and the true, the list value is removed from state and not restored

#### To Reproduce

```yaml
lowdefy: 3.23.2

pages:
  - id: test
    type: Context
    events:
      onInit:
        - id: init_state
          type: SetState
          params:
            visible: true
            object:
              list:
                - value: a
                - value: b
                - value: c
    blocks:
      - id: visible
        type: Switch
      - id: box
        type: Box
        visible:
          _state: visible
        blocks:
          - id: object.list
            type: List
            blocks:
              - id: object.list.$.value
                type: Paragraph
                properties:
                  content:
                    _state: object.list.$.value
```

Roshan931:
Fixes #1120

## Root cause

Hidden list's state array is deleted with no in-memory copy to restore from

## Changes

- Block now keeps an in-memory copy of a list's state value while the list is hidden (`captureHiddenValue` during `evaluate`) and republishes it when the list becomes visible again and the state field is absent (`restoreHiddenValue` at the start of `updateState`). Sub-block values are still republished on top of the restored array, and a `SetState` performed while hidden re-captures, so explicit changes keep precedence. Added three regression tests to `list.test.js` plus a changeset.

### DIFF
--- a/packages/engine/src/Block.js
+++ b/packages/engine/src/Block.js
@@ -339,6 +339,10 @@ class Block {
       repeat.value = true;
     }
 
+    if (this.isList() && !this.isVisible()) {
+      this.captureHiddenValue();
+    }
+
     if (this.visibleEval.output !== false) {
       this.propertiesEval = this.parse(this.properties);
       this.requiredEval = this.parse(this.required);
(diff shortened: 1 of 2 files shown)


Gervwyk:
Thank you — applied to `develop` via #2338 (cherry-picked with your authorship, plus a changeset), from where it flows into the v6 (`vite-hono`) and v7 (`auth-upgrade`) release lines. Closing this one since we release from `develop` rather than `main`.

PR #2323, opened 2026-08-23
fix(api): guard missing auth config

### LINKED ISSUE #2203: fix(api): missing null guard on `config.auth` causes typeerror crash instead of descriptive error
## Description

The `authorize` function destructures `auth` from `config` and then immediately accesses `auth.public` without checking if `auth` exists. If a request or endpoint config is missing the `auth` property (e.g., a misconfigured block), `auth` will be `undefined` and `auth.public` will throw an unhandled `TypeError: Cannot read properties of undefined (reading 'public')`. This crashes the API handler with an unhelpful stack trace instead of the descriptive `ConfigError` the code clearly intends to produce (as evidenced by the `ConfigError` thrown at the end of the function for invalid `auth.public` values). Both `authorizeRequest.js` and `authorizeApiEndpoint.js` call through this function, so any misconfigured config without an `auth` field will crash the entire request pipeline.

**Severity**: `high`
**File**: `packages/api/src/context/createAuthorize.js`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

mikemikimike:
Closes #2203

## Background

The API authorization helper accessed auth.public directly. Missing or null config.auth raised an unhandled TypeError instead of the intended descriptive ConfigError.

## Changes

- Guard auth.public accesses with optional chaining.
- Preserve the existing ConfigError message and config-key context.
- Add regression coverage for missing and null auth configuration.

## Compatibility

Valid auth.public values and role-based authorization behavior are unchanged.

## Verification

- `git diff --check` — passed.
- Docker: `node:24-alpine` with `pnpm install --frozen-lockfile --filter @lowdefy/api...`.
- `pnpm --filter @lowdefy/errors build` — passed.
- Focused Jest test — passed: 1 suite, 10 tests.

Command:

`node --experimental-vm-modules node_modules/jest/bin/jest.js --runTestsByPath src/context/createAuthorize.test.js`

No broader test suite was run because this change is isolated to the authorization context helper.

### DIFF
--- a/packages/api/src/context/createAuthorize.js
+++ b/packages/api/src/context/createAuthorize.js
@@ -30,15 +30,15 @@ function createAuthorize({ session }) {
 
   function authorize(config) {
     const { auth } = config;
-    if (auth.public === true) return true;
-    if (auth.public === false) {
+    if (auth?.public === true) return true;
+    if (auth?.public === false) {
       if (auth.roles) {
         return authenticated && auth.roles.some((role) => roles.includes(role));
       }
       return authenticated;
     }
     throw new ConfigError('auth.public must be true or false.', {
-      received: auth.public,
+      received: auth?.public,
       configKey: config['~k'],
     });
   }
--- a/packages/api/src/context/createAuthorize.test.js
+++ b/packages/api/src/context/createAuthorize.test.js
@@ -68,6 +68,12 @@ test('throws ConfigError with helpful message when auth.public is undefined', ()
   expect(() => authorize({ auth: {} })).toThrow('auth.public must be true or false.');
 });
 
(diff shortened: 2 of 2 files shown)


Gervwyk:
Thanks for the report and PR. We're not taking this one: the build stamps `auth` (`{ public, roles? }`) onto every page, request, endpoint and agent artifact, so `config.auth` can only be missing when a build artifact is corrupt or hand-edited — a bug, not a config mistake. Optional chaining there would turn a loud TypeError into a `ConfigError` that blames the app's `auth.public` config, which is misleading. Our rule is to fix at the source (the build) rather than guard at the consumer. Closing along with #2203.

PR #2334, opened 2026-08-28
fix(operators): Evaluate nested build operators under escaped prefixes

machielvdw:
## Summary

In the v6 build-time operator evaluator, four guards compared `operatorPrefix` for exact string equality with `'_build.'`. Nested build operators run under escaped prefixes (`'__build.'`, `'___build.'`, …) produced when `_function` re-parses its body, so those guards all took the runtime branch. A dynamic operator (e.g. `_function`) nested inside a `_build.function` body was deferred with a `~dyn` marker instead of evaluated at build time, and the raw operator object leaked into the built output — either failing the build loudly (`_array.concat must be evaluated on an array instance` when the leaked object reached the next fold iteration) or corrupting output silently (e.g. `__build.array.map` emitted verbatim into MongoDB aggregation pipelines). This was a regression against v4.5.2, whose BuildParser evaluated any operator under the current prefix unconditionally.

## Changes

- **@lowdefy/operators**: An `isBuildPrefix` helper (`/^_+build\.$/`) replaces the four exact `'_build.'` comparisons in `evaluateOperators` — the `~dyn` bubble-up skip, the dynamic-identifier deferral skip, the unknown-operator `ConfigError` path, and the dynamic-params deferral skip. Computed once per invocation since the prefix is constant per parse. Runtime prefixes (`'_'`, `'__'`, …) never match, so runtime deferral behaviour is unchanged.
- **@lowdefy/operators-js**: Integration regression tests in `function.test.js` — the dedupe-reduce reproduction with a nested `__build.function` comparator (including depth-3 `___build.*`) asserting the actual folded output, and the silent-corruption `__build.array.map`-in-`_build.function` shape, both run with the `dynamicIdentifiers` set the build passes (the ingredient the previous integration test was missing).

## Breaking Changes

- An unknown operator under an escaped build prefix (e.g. a typo like `__build.arrry.map` inside a `_build.function`) now raises a build `ConfigError` where it previously deferred silently into the output. This is the intended behaviour (it mirrors the canonical `_build.` prefix) but can surface pre-existing typos in real configs as new build failures. Noted in the changeset.

## Notes

- Unit tests cover each of the four guards individually under `'__build.'` and `'___build.'`, plus negatives: an escaped **runtime** prefix (`'__'`) still defers dynamic identifiers, and a runtime operator written inside a build-prefix parse passes through untouched — the fix does not make runtime operators evaluate at build time.
- `_operator` (also in the dynamic set) takes a `name` param and dispatches directly without re-deriving prefixes, so its behaviour under build prefixes is unchanged and matches v4.
- The only other prefix-derived logic in the tree is `walker.js`'s `startsWith('_build.')` config-level detection, which is correct as-is: escaped keys are only meaningful inside a function-body parse.
- Test runs after rebuilding `@lowdefy/operators`: operators 212 ✓, operators-js 1032 ✓, build 1718 ✓, api 470 ✓.

### DIFF
--- a/.changeset/fix-operators-escaped-build-prefix.md
+++ b/.changeset/fix-operators-escaped-build-prefix.md
@@ -1,5 +1,5 @@
 ---
-'@lowdefy/operators': patch
+'@lowdefy/operators': minor
 ---
 
 fix(operators): Recognise escaped build prefixes in evaluateOperators guards.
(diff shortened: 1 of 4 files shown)


machielvdw:
0.0.0-experimental-20260828081645

Gervwyk (approved):

PR #2345, opened 2026-08-29
fix: Accept a repeated MCP-Protocol-Version header on /api/mcp

Gervwyk:
## Summary
Claude Code (2.1.x) fails to connect to a Lowdefy MCP server with **"MCP endpoint not found"** even though the route exists. Root cause, reproduced against `@hono/mcp@0.3.0` + SDK 1.29:

- Claude Code builds request headers as `new Headers({ ...commonHeaders, ...requestInit.headers })`, where `commonHeaders` carries the negotiated `mcp-protocol-version` and `requestInit.headers` (its "init-projection headers") carries a static `MCP-Protocol-Version`. The Fetch `Headers` constructor treats the two spellings as one header and **appends**: the server receives `mcp-protocol-version: 2025-11-25, 2025-11-25`.
- `@hono/mcp` `#validateProtocolVersion` compares that raw string against `SUPPORTED_PROTOCOL_VERSIONS` and answers **404** `Unsupported protocol version` on every non-`initialize` request (it handles an array, not a comma-joined string). Claude Code renders any 404 from the endpoint as "MCP endpoint not found".

Fix: `normalizeMcpProtocolVersionHeader` runs at the top of `mcpHandler` in both servers and collapses a *repeated* value to the single version it names (`c.req.header()` reads the raw request headers live, so the transport then sees one value). A header naming two **different** versions is left untouched for the transport to refuse.

## Test plan
- [x] Repro script: `2025-11-25` → 200; `2025-11-25, 2025-11-25` → 404 `Unsupported protocol version`; `new Headers({'mcp-protocol-version': …, 'MCP-Protocol-Version': …}).get(…)` → `"2025-11-25, 2025-11-25"`.
- [x] New `normalizeMcpProtocolVersionHeader.test.mjs` in both servers (repeated value, both-spellings construction, single value, two different values, absent header); `routes/mcp` suites pass (18/18 each).
- [ ] Claude Code `claude mcp add --transport http … /api/mcp` connects and lists tools once a pin carrying this ships (encircle-mvp will verify).

### DIFF
--- a/.changeset/fix-mcp-protocol-version-header.md
+++ b/.changeset/fix-mcp-protocol-version-header.md
@@ -1,0 +1,8 @@
+---
+'@lowdefy/server': patch
+'@lowdefy/server-dev': patch
+---
+
+fix: Accept a repeated `MCP-Protocol-Version` header on `/api/mcp`.
+
+Claude Code sends the negotiated protocol version under two spellings of the header, which the Fetch Headers constructor folds into one comma-joined value (`2025-11-25, 2025-11-25`). The MCP transport compared that raw string against its supported list and answered 404 "Unsupported protocol version" — shown by Claude Code as "MCP endpoint not found". The route now collapses a repeated value to the single version it names before the transport reads it; a header naming two different versions is still refused.
--- a/packages/servers/server-dev/src/routes/mcp.js
+++ b/packages/servers/server-dev/src/routes/mcp.js
@@ -19,6 +19,7 @@ import { createMcpServer, getMcpResourceMetadataUri, getMcpResourceUri } from '@
 import { type } from '@lowdefy/helpers';
 
 import authJson from '../../lib/build/auth.js';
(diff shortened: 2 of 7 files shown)


Gervwyk:
Correction to this PR's description, for the record. The `#validateProtocolVersion` error seen in encircle-mvp's dev log was **not** caused by a comma-joined duplicate header. Logging the raw header on a live server showed Claude Code 2.1.251 *probing* with `MCP-Protocol-Version: 2026-07-28` (newer than any revision `@modelcontextprotocol/sdk` 1.29/1.30 supports) on its first POST, then re-initialising and continuing with `2025-11-25` — expected negotiation. The visible damage was our error handler turning the transport's 404 into a logged internal error and a 500; that is fixed in #2353. The collapsing of a repeated header value added here is still correct (Node joins duplicate request headers with `, `, and the `Headers` constructor folds the two spellings), but it was not the trigger of that trace. The "MCP endpoint not found" in Claude Code that motivated this PR turned out to be the dev server running from a checkout without the `/api/mcp` route.

PR #2366, opened 2026-09-04
fix(ai): a failed agent turn no longer poisons the conversation

Gervwyk:
## Problem

Reported from a production app: after one failed agent turn, every later message in the chat produced an error toast containing the entire conversation serialised as JSON.

The AI SDK client pushes the assistant message on the stream's `start` chunk, before any content arrives. A request that fails after that point leaves an assistant message with `parts: []` in the client's history. On the next send that shell fails UIMessage validation (`Message must contain at least one part`). The `TypeValidationError` message embeds `JSON.stringify(value)`, i.e. the whole conversation, and that text reached the client verbatim because validation ran outside the handler's try/catch and `createUIMessageStream` used the default `onError`. Every subsequent turn failed the same way until a page reload.

## Fix

- **AgentChat** (`blocks-antd-x`): on `onError`, drop assistant messages with no parts so the user can simply retry.
- **handleAgentChat** (`ai-utils`):
  - ignore empty messages before validation (nothing in them for the model);
  - run validation inside the same try as the stream read so the fault is logged server-side (`Agent stream failed.`);
  - pass an `onError` to `createUIMessageStream` that maps `TypeValidationError` to a short client message instead of the value dump. Other errors keep their message.

## Tests

`@lowdefy/ai-utils`: 176 passed (3 new: empty messages dropped, validation failure logged + written as a stream error, `onError` redaction).

https://claude.ai/code/session_01QiqYjqaQEmWiYu9kDxBbVZ

### DIFF
--- a/.changeset/agent-chat-error-empty-assistant.md
+++ b/.changeset/agent-chat-error-empty-assistant.md
@@ -1,0 +1,6 @@
+---
+'@lowdefy/ai-utils': patch
+'@lowdefy/blocks-antd-x': patch
+---
+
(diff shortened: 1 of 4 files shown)


Gervwyk:
Retargeted to v7 (the line the Encircle app pins): superseded by the v7 PR.

PR #2375, opened 2026-09-08
feat(auth): magic-link landing page keeps the token unconsumed until a real click

Gervwyk:
## Problem

Magic-link sign-in emails link straight at Better Auth's `GET /api/auth/magic-link/verify`, which consumes the single-use token atomically on the first GET (`allowedAttempts` is explicitly ignored upstream). Corporate mail security (Microsoft Defender Safe Links, Proofpoint URL Defense, Mimecast) pre-fetches links at delivery time, so the scanner spends the token and the person clicking lands on `INVALID_TOKEN`. The server already short-circuits HEAD prefetches; the GET prefetch was the remaining gap.

## Fix

- **`auth.authPages.magicLink`** (opt-in page path). When set, the engine points the sign-in email at that Lowdefy page instead of the verify endpoint, copying the whole verify query (token, `callbackURL`, `newUserCallbackURL`, `errorCallbackURL`) onto it. Origin is taken from the verify URL so the zero-config path (no `BETTER_AUTH_URL`) keeps working. Unset, behaviour is byte-for-byte unchanged.
- **`MagicLinkVerify` action** (actions-core + engine + client): on a real click it navigates the browser to the verify endpoint. Token and destinations default from the landing page's URL query; params override. `callbackUrl: false` is rejected (redirect hop). Chain stops after navigation like every other navigating auth method.
- **Build validation**: the page is not required when magic link is enabled, so existing apps keep building. It is also allowed when magic link is off: modules contribute `authPages` roles statically (they cannot read the auth projection inside the auth block), so a stricter rule would break every consumer of a module that ships the landing page.
- **MagicLinkEmail**: the URL is also shown as readable text under the button for mail clients that strip or rewrite buttons.
- **Docs**: new `MagicLinkVerify` action page, a "Magic link sign-in" section in Login and Logout, and "The magic link landing page" in Auth configuration, all stating the page must bind the action to a click and never auto-redirect (JS-executing sandboxes would consume the token anyway). Agent docs regenerated for the touched pages only; `pnpm docs:content` on v7 also shows unrelated pre-existing drift that was left out of this PR on purpose.

## Tests

`@lowdefy/api`, `client`, `actions-core`, `engine`, `email-templates`, `build` suites pass, and the Vercel docs deploy check has been failing on every recent v7 commit independently of this PR (build snapshot fixtures unchanged; the cross-module fixtures time out at 5 s when two jest runs share the machine, and pass in isolation).

## Follow-up

A sibling PR adds `auth.emailOTP` and carries a one-time code in the same email for people whose mail security also blocks the landing page.

### DIFF
--- a/.changeset/magic-link-landing-page.md
+++ b/.changeset/magic-link-landing-page.md
@@ -1,0 +1,30 @@
+---
+'@lowdefy/actions-core': patch
+'@lowdefy/email-templates': patch
+'@lowdefy/engine': patch
+'@lowdefy/client': patch
+'@lowdefy/build': patch
+'@lowdefy/api': patch
+---
+
+feat(api,client,build,actions-core): magic-link landing page keeps the token unconsumed until a real click
+
+A magic-link sign-in token can only be used once, and the auth server spends it on
+the first request to the link — including a request nobody made. Corporate email
+security (Microsoft Defender Safe Links, Proofpoint URL Defense, Mimecast) opens
+every link in a message as it is delivered, so by the time the person clicks, the
+link is already used up and they land on the error page.
+
+Set the new `auth.authPages.magicLink` to a page in your app and the sign-in email
+links there instead, carrying the token and the sign-in destinations along. The
+page just sits there when a scanner fetches it; a button on it runs the new
+`MagicLinkVerify` action, which finishes the sign-in. The token is only spent when
+a person clicks, so scanned mailboxes can sign in again.
+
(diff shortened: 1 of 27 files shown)


Gervwyk:
Screenshots: the sign-in email with the readable link under the button, the new MagicLinkVerify docs page, and the new sections in Login and Logout and Auth configuration.

Magic-link email with readable link fallback

MagicLinkVerify action docs

Login and Logout: Magic link sign-in section

Auth configuration: the magic link landing page

PR #2383, opened 2026-09-08
feat: backport the v7 developer-experience tooling and fixes to v6

Gervwyk:
## Summary

Ports the non-auth developer-experience work and fixes from `v7` to `v6`. The `v6` and `v7` branches share the Hono + Vite architecture; the difference between them is the BetterAuth upgrade. Everything auth-related (BetterAuth, organizations, tenant wall, two-factor, passkeys, captcha, phone/email OTP, magic-link landing, OAuth for MCP) stays on `v7`. Every commit here is cherry-picked with `-x` from `v7`, adapted to v6's Auth.js user model where the v7 version leaned on v7-only auth concepts.

## What lands on v6

**Dev MCP / headless agent tooling (`@lowdefy/server-dev`)**
- `lowdefy_run_journey`: drive a page through declarative steps (grid cell buttons, confirm dialogs and menu items included), plus `/lowdefy-docs/journey`.
- `lowdefy_run_endpoint`: run an Api endpoint routine headlessly. With `system: true` (#2399) it runs as a system context through the api package's `runDetachedEndpoint`, the entry point `/api/detached` uses: no user, endpoint `auth` not checked, `InternalApi` allowed. This is the local test path for scheduled and detached-only routines without `CRON_SECRET`; `system` plus `user` is a 400, and nested detached `CallApi` steps are not faked.
- `lowdefy_restart`, and the dev manager watches local plugin server-side sources.
- Contextual hazards from the dev MCP (schemas, docs, config lookups; `hazards.json` ships in `@lowdefy/docs-content`).
- Build, restart and client/server error events pushed to agents over `/lowdefy-docs/events`.
- Server errors collected into `build_status`; config source carried to MCP callers.
- Docs and MCP responses marked stale while the build is failing.
- `run_request` runs as a named caller through the `user` param (resolved through v6's dev-session pipeline).
- Headless pages open with a `urlQuery`; headless capture waits for the page's async lifecycle; the browser context closes when `openPage` fails mid-open.
- The dev manager refuses to start twice for the same app.

**CLI**
- `lowdefy test` runs `tests/journeys/*.yaml` against a dev server, with docs on the new Config tests page.
- swc's test-file exclusion escapes its dots so `commands/test/test.js` compiles.

**Build**
- Prod-gated build warnings are badged "fails in prod" in dev.
- An operator at a block's `class` is evaluated at runtime.
- `_build.env` warns on every build when it inlines an unset variable (new `secrets` check slug).
- Every plugin subpath resolution failure is treated as an absent optional module.
- The agents, notifications, websockets and mcp builders collect every error instead of stopping at the first; deprecated styles are validated in one place.
- `auth.dev` alone is no longer an auth stack, and the dev mock user reaches `_user` in the browser.

**Api / operators / layout**
- `:while` routine control; `:reject` documented as a reply, not an exception.
- `_type` gains an `empty` test.
- `layout.align` is applied instead of silently dropped.

**Errors**
- MongoDB driver errors become `ServiceError` with a hint; the raw driver message stays in the server terminal and off the wire.
- `AuthorizationError` (403) joins `AuthenticationError` (401): v6's request, endpoint, agent and websocket gates now answer an anonymous caller with 401 and a wrong-roles caller with an opaque 403, logged at warn level, instead of a masked 500. All three servers' error handlers map both, and hono `HTTPException` answers are sent as-is.
- Error classes applied consistently in runtime plugins; `UserError` is matched by name across package boundaries; plugin-thrown Lowdefy errors get their `configKey` from the interface layer.

**Servers**
- Unserved `/.well-known/*` paths answer 404 instead of the app shell.
- One same-origin guard for browser POST routes.

**Blocks**
- antd 6 deprecated props removed (Drawer, Selector, MultipleSelector, PhoneNumberInput, Collapse…); QRCode gallery uses `layout.align`.
- `DangerousHtml` warns when `DOMPurifyOptions` change after mount.
- Blocks declare the events and properties their components already use (AgGrid, Tabs, DropdownButton, EChart, loaders…).

## Deliberately left on v7

Everything BetterAuth: `plugin-better-auth`, organizations and the tenant wall, two-factor and passkeys, captcha and phone number auth, email OTP, magic-link landing page, MCP OAuth resource server and organization selection, the caller normalisation (`normalizeCaller`, snake_case caller fields), `createSystemContext`, `TwoFactorEnrolmentRequiredError`, the auth reference apps and migration fixture, and the v7 build fixtures that project `_build.authConfig`.

Already on v6 through `develop` and skipped as empty: the color-scheme sync, the JIT page-before-request fix, the tiptap mention id, AgentChat attachments, the helpers dot-path grammar, the nunjucks LRU, the pino startup logging, the client-bound error redaction, and MCP serverInfo branding.

## Behaviour changes worth a look

- A protected request/endpoint/agent/websocket called without a session now returns 401 (`AuthenticationError`) and a caller with the wrong roles 403 (`AuthorizationError`, message unchanged), where v6 previously surfaced a 500 with the masked "does not exist" message.
- An `auth` block whose only key is `dev` (or an empty `auth: {}`) now builds with `auth.configured: false`, the same as v7.

## Verification

- `pnpm build` clean on the merged branch.
- Full workspace test run (`pnpm -r --no-bail test`, plus the root-excluded `connection-mongodb` and `plugin-aws`): every package green, 7,600+ tests.
- Live check of `run_endpoint` with `system: true` against an `InternalApi` endpoint declaring `schedules`: the system run executes and returns; the same call without `system` is still refused; `system` plus `user` answers 400.
- Live smoke test of `lowdefy dev` on the docs app: `/lowdefy-docs`, `build-status` (carrying the new `secrets` warning), `page-config`, the MCP `tools/list` with `lowdefy_run_journey`, `lowdefy_run_endpoint` and `lowdefy_restart`, a headless journey run, `run_request` user validation, and the `/.well-known` 404.
- Snapshot fixtures under `packages/build/src/tests` regenerated on v6 (v7-only auth fixtures dropped).

https://claude.ai/code/session_01FvHs4eMUb8pZw6PvxAzfNm


machielvdw:
0.0.0-experimental-20260914123122

machielvdw:
0.0.0-experimental-20260914131506

Issue #2397, opened 2026-09-10
AgentConversations: mark a row as loading


JohannMoller:
## Problem

`AgentConversations` exposes `activeKey` to mark the selected row, but there is no way to show that the selected conversation is still being loaded. When an app fetches a transcript on select, the row flips to selected immediately and nothing indicates that the read is in flight; the only loading signal has to live in the chat column.

## Proposal

Either `properties.loadingKey: string` (the row currently loading shows a small spinner in place of its icon / at its end), or `items[].loading: boolean` on the item schema. `loadingKey` mirrors `activeKey` and is the smaller addition.

## Context

Requested as part of lowdefy/modules-mongodb#221; not deliverable from config today.

Issue #2396, opened 2026-09-10
AgentChat: external messages sync can skip a conversation with the same count and last id


JohannMoller:
## Problem

The external-messages sync in `AgentChat.js` only calls `setMessages(msgs)` when the incoming array's length or last message id differs from the previous one (`prevExternalRef = { count, lastId }`). That memo is not reset when `conversationId` changes.

Two conversations whose transcripts have the same number of messages and the same last message id therefore do not re-sync when switching from one to the other. This is not hypothetical: every assistant message persisted by the onFinish hook has id `""` (the AI SDK ids the assistant reply `""` unless `generateMessageId` is passed — see the transcript-restore note in lowdefy/modules-mongodb `apps/demo/e2e/ai-reporting/chat-transcript-restore.spec.js`), so any two two-turn conversations collide on `count: 2, lastId: ""`. The switch shows an empty transcript.

Apps currently avoid it only by chance: blanking `messages` to `[]` before setting the new array flips the count through 0.

## Proposal

Either reset `prevExternalRef` in the `conversationId`-change effect, or include `effectiveConversationId` in the comparison key. Ideally both this and the assistant-id `""` problem are addressed (pass `generateMessageId` to `useChat`/the transport so persisted assistant messages have unique ids).

## Context

Found while reviewing lowdefy/modules-mongodb#223.

JohannMoller:
Correction: the assistant-id root cause is already fixed on `main` — `handleAgentChat.js` passes `generateMessageId` to both stream paths. Transcripts persisted before that still carry `''`, so the sync-memo fix stands; see the linked PR.

Issue #2395, opened 2026-09-10
AgentChat: add a loading property (transcript skeleton + disabled composer)


JohannMoller:
## Problem

`AgentChat` has no loading or disabled state. An app that loads a conversation's transcript from its own store (the `messages` property) has nothing to show between selecting a conversation and the read resolving: the transcript is empty and the composer is live, so a message can be sent into a conversation whose history has not synced yet — and is then persisted under a transcript the block has not seen.

The only config-level workaround is to hide the whole block with CSS (`display: none`, since unmounting would drop the `useChat` instance) and render a `Skeleton` beside it. That takes the composer out of the DOM rather than disabling it, uses the app's layout rather than the block's, and has to be sized to match the block by hand.

## Proposal

- `properties.loading: boolean` — when true, render a skeleton (a few bubbles' worth) in the message area in place of the transcript and set the `Sender` to disabled. The welcome and suggestions are hidden while loading.
- Optionally, independent of loading, `sender.disabled: boolean` for apps that need to gate the composer for other reasons.

Both are reactive properties like `messages` and `suggestions`, so an app can drive them from state.

## Context

Found while fixing lowdefy/modules-mongodb#221. The module-side workaround (hidden block + sibling skeleton) is in `modules/ai-reporting/pages/chat/components/chat_workspace.yaml` in lowdefy/modules-mongodb#223.

PR #2398, opened 2026-09-10
fix(blocks-antd-x): AgentChat stops the stream on switch, re-syncs same-shaped transcripts, adds loading

### LINKED ISSUE #2394: AgentChat: abort the in-flight stream when conversationId changes
## Problem

`AgentChat` clears its messages when `conversationId` changes (`packages/plugins/blocks/blocks-antd-x/src/blocks/AgentChat/AgentChat.js`, the `prevConversationIdRef` effect: `setMessages([])`), but it does not abort a reply that is still streaming. The stream keeps running and its chunks continue to land in `useChat`'s message list — which now belongs to the new conversation. Switching away from a mid-stream conversation therefore leaks partial assistant content into the one being opened.

Consumers can only work around this by calling the block's `stop` method from config before every switch (`CallMethod { blockId, method: stop }`), which has to be remembered at every switch path and depends on the block being mounted when the action runs.

## Proposal

In the `conversationId`-change effect, call `stop()` before `setMessages([])`. `stop()` on an idle chat is a no-op, so this is safe unconditionally.

## Context

Found while fixing lowdefy/modules-mongodb#221 (ai-reporting conversation switching). The module-side workaround is `modules/ai-reporting/pages/chat/actions/stop_stream.yaml` in lowdefy/modules-mongodb#223.

JohannMoller:
Closes #2394, #2395, #2396.

## Changes (`AgentChat.js`)

**Stop the stream on a `conversationId` change (#2394).** The id-change effect cleared messages but left the stream running, so a reply still streaming into the conversation being left kept appending into the new one. `stop()` now runs before `setMessages([])`; it is a no-op on an idle chat.

**Re-sync external messages across a switch (#2396).** The `messages` sync skipped `setMessages` when the array had the same length and last id as before, and that memo did not reset on a switch. Every assistant message an older onFinish hook persisted carries id `''` (the root cause is fixed on `main` by `generateMessageId` in `handleAgentChat`, but stored transcripts keep the shape), so any two same-length transcripts collided and a switch showed an empty chat. The conversation id is now part of the comparison key.

**`loading` property (#2395).** While `true`, a skeleton replaces the message area, suggestions are hidden and the `Sender` is `disabled`, so nothing can be sent into a conversation whose history has not synced yet. For apps that read a transcript from their own store on select and drive `messages` from state.

## Also
- `schema.json`: `loading` declared.
- Docs (`agents/agentchat.md`): `loading` row; `conversationId` and `messages` rows say what now happens on a switch.
- e2e: a `chat_switch` fixture with two same-shaped transcripts (2 messages, assistant id `''`) and a spec asserting the switch swaps them both ways; a `chat_loading` fixture and spec asserting the skeleton, hidden suggestion and disabled textarea, and that the transcript returns when loading clears.
- Changeset: `@lowdefy/blocks-antd-x` minor.

## Verification
Edited in a sparse checkout without the workspace installed, so I have not run the blocks-antd-x e2e suite here — the two new specs need a CI run. Source and fixture were checked with prettier against the repo config. The module-side workarounds these replace are in lowdefy/modules-mongodb#223, where the equivalent behaviour is covered by e2e.

Not included: #2397 (AgentConversations `loadingKey`) is a different block.


JohannMoller:
## Self-review

**Blocking: the stop-on-switch fix does nothing as written.** The block keys `useChat` by `id: effectiveConversationId`. In `@ai-sdk/react` 3.0.178 the hook recreates its `Chat` instance synchronously during render when the id changes, and the `stop` it returns is `chatRef.current.stop`, i.e. the *new* instance's method. The effect that calls `stop()` runs after that render, so it stops the fresh, idle Chat (and `Chat.stop` early-returns unless that instance is streaming/submitted). Nothing aborts the request that was actually in flight.

The consequence is also different from what the PR body claims. The hook re-subscribes to the new Chat, so the old stream's text does **not** append into the new message list. What leaks is the callbacks: `onData`, `onToolCall`, `onFinish` and `onError` are routed through a ref to the latest callbacks, so the abandoned stream keeps firing `onDataPart` and tool-call events into the page (a `dataParts` consumer still receives the old conversation's parts under the new id), and the server keeps generating a reply nobody sees.

Fix: stop the instance being left, not the one being entered. A cleanup effect keyed only on the id closes over the previous render's `stop`:

```js
useEffect(() => () => stop(), [effectiveConversationId]);
```

The e2e switch test cannot catch this: it only covers the re-sync case and no fixture has an in-flight stream.

**Non-blocking:**

- `prevExternalRef` starts with `conversationId: null`, so the first sync effect run always calls `setMessages(msgs)` even when `messages` is undefined, which the comment says means "no sync". Harmless on an empty Chat, but seed the ref with the initial id.
- `!isLoading` on the suggestions row is redundant with `!isEmpty` after a switch (messages are cleared), only matters if `loading` flips without an id change. Kept as belt and braces.
- If an app flips `loading` true mid-stream, the Sender is both `loading` and `disabled` and the stop button is unreachable. `loading` is meant for the pre-sync window; docs should say so.
- The e2e specs were not run locally before the rebase either.

Fixes for the blocking item and the first and third non-blocking items follow in the next commit.

